PR #115, opened 2018-11-30
Fix invalid settings file path and make parseSettingsFile more generic

walkre-niboshi:
- Fix invalid setting file path (`"~/.moerc.toml"`  to `"~/.moerc.toml".expandTilde`)
- Rename "config" to "settings" to be consistent naming
- Make parseSettingsFile more generic


fox0430:
Good. Thanks.

PR #116, opened 2018-11-30
Make settings file location XDG compliant

SolitudeSF:
Maybe filename could be changed to more generic `config.toml`, since its now located in separate folder.


fox0430:
Good idea! Thank you.

PR #139, opened 2019-02-22
Fix #138:   Use my forked nim-ncurses repository instead of origin temporarily

### LINKED ISSUE #138: Can not install and build
It is because the ncurses package can't install.

walkre-niboshi:
And I fixed some type conversions in ui.nim.
However this fix is foolishly honest, so I have no confidence ... .


fox0430:
@walkre-niboshi Can i marge this PR ?

PR #240, opened 2019-04-26
add: D, m, ', line-directed G. fix: new files

jrfondren:
D - delete text to end of line
  m - set a mark (no argument for now)
  ' - go to mark (no argument for now)
  G - 1G should go to line 1, 10G to line 10, etc.
      without arguments, it goes to the end of the file
      since this distinguishes between 1 and no argument,
      used origCmd. which is ugly but minimal change.

fix:

  if you run "moe new-file-name", and then try to save with :w, moe
  complains that you haven't told it what file name to save as.
  moe should default to 'new-file-name' the way it does when that file
  already exists and moe is started to edit it.

still wanted:
  ZZ - save and quit
  zt, zb, z., zz - move display so current line is first visible line,
                   etc.
  ^ how are multiple-key commands handled?

  typing ] should move the cursor past a ] that was added by a
  previously typed [, etc.
  ^ I thought this might require marking the chars somehow, but ] could
    probably just always skip past ] when in a mod where [ are auto entered.

  undo, even 1-stack-deep undo!
  with undo I can use moe for practical tasks

### DIFF
--- a/src/moe.nim
+++ b/src/moe.nim
@@ -50,7 +50,7 @@ proc main() =
         status.bufStatus.add(BufferStatus(filename: "".toRunes))
         status.bufStatus[0].buffer = newFile()
     else:
-      status.bufStatus.add(BufferStatus(filename: "".toRunes))
+      status.bufStatus.add(BufferStatus(filename: parsedList.filename.toRunes))
       status.bufStatus[0].buffer = newFile()
   else:
     status.bufStatus.add(BufferStatus(filename: "".toRunes))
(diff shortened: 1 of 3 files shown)


jrfondren:
Will update after the v0.0.7 merge

fox0430:
@jrfondren 

Thank you feedback and PR!!  

Will change to use new-file-name with default file name.

ZZ, zt, zb etc will be implemented soon.
The multiple key is waiting again when the first key is pressed.

`Will change the behavior when the '[' key is entered.

Undo/redo will be implemented soon!

jrfondren:
Thanks, I'll look forward to it.

I don't mean that 'new-file-name' should be a default, but that if you start moe with a filename that doesn't exist, that filename should be the default. right now, if the user does "moe mycode.nim" -> change something -> ":w" -> (if mycode.nim already existed, then the :w succeeds; otherwise, this is an error)

fox0430:
@jrfondren 

Ok. I understand.
It's a bug and will fix it.

Thank you!

Issue #260, opened 2019-04-29
Window width gets smaller on each refresh


jrfondren:
This is easier to observe with very long wrapping lines, but even a file containing just 11 chars in a 112-column terminal is enough:
```
xxxxxxxxxxx
```

### replication
open the file with those x'ses, and
- hold Ctrl-L
or:
- alternate holding left and right arrows
eventually this will happen to the display:
```
xxxxxxx
xxxx
```
->
```
xxxxx
xxxxx
x
```
->
```
xxx
xxx
xxx
xx
```
->
```
x
x
x
x
x
x
x
x
x
x
x
```
->
and then nothing's displayed at all.

It might take a few seconds of refreshing the display for the bug to occur, but moe immediately starts breaking down on a file with a 239-wide line.

jrfondren:
I traced this for a bit and found that the following code was shrinking the display, in src/moepkg/editorview.nim:

```
  if widthOfLineNum != view.widthOfLineNum: view.resize(buffer, view.height, view.width+view.widthOfLineNum-widthOfLineNum, widthOfLineNum)
```
^ widthOfLineNum is always 2 (in tests), but view.widthOfLineNum is always 0, so on each update the view shrinks by 2:

```
resizing: 110, 2, 0
resizing: 108, 2, 0
resizing: 106, 2, 0
resizing: 104, 2, 0
resizing: 102, 2, 0
resizing: 100, 2, 0
resizing: 98, 2, 0
resizing: 96, 2, 0
resizing: 94, 2, 0
```

I tried setting view.widthOfLineNum in this proc, but it was still always zero! Something was resetting it. Looking around, I noticed this line:

```
src/moepkg/editorview.nim:  view.widthOfLineNum = if lineNumber: buffer.len.numberOfDigits+1 else: 0
```

And then I realized: I'd turned off line-numbers in my moe config, in moerc.toml. I tried turning them back on and confirmed that the bug was no longer present. The view only shrinks with this configuration:

```
[Standard]
number = false
```

So the fix seems to be to add a `lineNumber and` to the test in proc update.

PR #267, opened 2019-05-07
EditorView.update shouldn't resize display to make room for line numb…

jrfondren:
…ers if line numbers aren't displayed

fix for #260


fox0430:
@jrfondren Thank you! merge into develop branch.

PR #285, opened 2019-05-14
Support change color setting

fox0430:
#184


fox0430:
TODO: Fix filer mode color

PR #394, opened 2019-09-07
Support clip board

### LINKED ISSUE #392: Add support for copy & paste
Copy & paste (by all 3 means: keyboard, mouse, command line) seems essential for good interoperability.

As inspiration how to achieve that on "all" platforms, refer to https://github.com/neovim/neovim/blob/master/runtime/autoload/provider/clipboard.vim .

fox0430:
Close #392

### DIFF
--- a/src/moepkg/normalmode.nim
+++ b/src/moepkg/normalmode.nim
@@ -351,6 +351,7 @@ proc sendToClipboad*(registers: Registers, platform: Platform) =
   case platform
     of linux: discard execCmd("echo " & $buffer & " | xclip")
     of wsl: discard execCmd("echo " & $buffer & " | clip.exe")
+    of mac: discard execCmd("echo " & $buffer & " | pbcopy")
     else: discard
 
 proc yankLines(status: var EditorStatus, first, last: int) =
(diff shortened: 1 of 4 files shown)


fox0430:
## TODO
Supports Windows, WSL, macOS

fox0430:
Put Windows support on hold.

src/moepkg/normalmode.nim
```diff
@@ -346,6 +346,17 @@ proc deleteCharacterBeginningOfLine(bufStatus: var BufferStatus) =
   bufStatus.expandedColumn = 0
   for i in 0 ..< beforColumn: deleteCurrentCharacter(bufStatus)
 
+proc sendToClipboad*(registers: Registers, platform: Platform) =
+  let buffer = if registers.yankedStr.len > 0: $registers.yankedStr else: $registers.yankedLines
+  case platform
+    of linux:
+      ## Check if X server is running
+      let (output, exitCode) = execCmdEx("xset q")
+      if exitCode == 0: discard execShellCmd("echo " & $buffer & " | xclip")
```
dumblob: Please don't use `echo` here as it's not clearly defined in POSIX what it can or can't. All `echo`s I know about interpret some backslash sequences (each `echo` implementation a different set of them :cry:), some ignore/trim some characters, etc. Generally use `printf` instead of `echo` everywhere as it's defined in POSIX :wink:.

In this case though, I'd recommend using the following POSIX-compliant variant (which should be also slightly faster as it doesn't require the shell to fork itself like the current `|` pipe): `"xclip <<END\n" & $buffer & "\nEND\n"` (feel free to change `END` to something else/longer to avoid it's existence in `$buffer` or even check beforehand whether it's a substring of `$buffer` and generate randomly a different delimiter string).

fox0430: Thank you advice!
I will fix it.

PR #419, opened 2019-10-25
fix #417

### LINKED ISSUE #417: Sometimes ":bd n" Fails and moe Crashes with "sysFatal"
When we execute ``:bd n`` where ``n`` is more than buffers count,
moe crashes with ``Error: unhandled exception: index 1 not in 0 .. 0 [IndexError]``

loloiccl:
fix #417


fox0430:
@loloiccl 
Thank you very much!

PR #484, opened 2020-01-22
Fix #473

### LINKED ISSUE #473: "sh line 2: warning: here-document at line 0 delimited by end-of-file (wanted `MWexnsqihH')" displayed when deleting characters continuously

fox0430:
Close #473

### DIFF
--- a/src/moepkg/normalmode.nim
+++ b/src/moepkg/normalmode.nim
@@ -430,7 +430,14 @@ proc genDelimiterStr(buffer: string): string =
     if buffer != result: break
 
 proc sendToClipboad*(registers: Registers, platform: Platform) =
-  let buffer = if registers.yankedStr.len > 0: $registers.yankedStr else: $registers.yankedLines
+  var buffer = if registers.yankedStr.len > 0: $registers.yankedStr else: $registers.yankedLines
+  if buffer.len < 1: return
+
+  if buffer in "`":
+    for i in 0 ..< buffer.len:
+      if i == 0 and buffer[0] == '`': buffer = "\\" & buffer
+      elif buffer[i] == '`': buffer.insert("\\", i - 1)
+
   let delimiterStr = genDelimiterStr(buffer)
   case platform
     of linux:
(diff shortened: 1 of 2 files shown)


walkre-niboshi:
Following steps could still cause this issue.

1. Start moe.
1. Enter 'i' to switch to insert mode.
1. Enter '`' (backquote).
1. Press the ESC key to switch to normal mode.
1. Enter 'x'

fox0430:
@walkre-niboshi 
Thank you.
I will fix it.

src/moepkg/normalmode.nim
```diff
@@ -430,22 +430,24 @@ proc genDelimiterStr(buffer: string): string =
     if buffer != result: break
 
 proc sendToClipboad*(registers: Registers, platform: Platform) =
-  var buffer = if registers.yankedStr.len > 0: $registers.yankedStr else: $registers.yankedLines
+  let buffer = if registers.yankedStr.len > 0: $registers.yankedStr else: $registers.yankedLines
   if buffer.len < 1: return
 
-  if buffer in "`":
+  var sendStr = ""
```
walkre-niboshi: Maybe you can make the code shorter as below.
```
let sendStr = buffer.replace("`", "\\`")
```

walkre-niboshi:
Let me present another example which involves with viaual mode.

1. Start moe, redirecting stderr to a file (for example `./moe 2> error.txt`).
1. Enter 'i' to switch to insert mode.
1. Enter '\' and '`'.
1. Press the ESC key to switch to normal mode.
1. Move to the first character('\').
1. Enter 'v' to switch to visual mode.
1. Move to the last character.
1. Enter 'x' to delete the all characters.
1. Exit moe, and see the file you redirected stderr.

In my environment, by following the above steps I can see the error: `sh: 4: Syntax error: EOF in backquote substitution`.

By the way

walkre-niboshi:
I suggest that we quote the the delimiting identifier of the here document.

walkre-niboshi:
In the first place, we shouldn't use a procedure which just executes a command such as `osproc.execProcess proc` instead of `os.execShellCmd` ?

fox0430:
## TODO
Add unit test

PR #551, opened 2020-03-23
Fix #550

### LINKED ISSUE #550: Ex mode does not exit from previous buffer when opening new buffer with e command

fox0430:
Close #550


fox0430:
## TODO
Add unit test

PR #613, opened 2020-04-30
Highlights "TODO" in comments #461

jiro4989:
Add highlighting TODO and WIP.
I will remove `WIP` if you don't need WIP.
(WIP is a word for my test code)


fox0430:
@jiro4989 
Excellent! thank you!
```WIP``` is also welcome!

PR #622, opened 2020-05-17
Embed version from .nimble file into cli usage message at compile time.

quantimnot:


quantimnot:
I added an integration test for this. I understand if you don't want it though.
It uses [shpec](https://github.com/rylnd/shpec) as the test runner.
Some shell integration tests are best run with a terminal session manager.
I configured the shpec test suite to use an existing session manager or to skip tests otherwise.
I only configured [abduco](https://github.com/martanne/abduco ), but more can be added.

fox0430:
@quantimnot 

Great! I wanted this. Thank you very much.

PR #623, opened 2020-05-17
Fixes breaking changes from https://github.com/nim-lang/Nim/pull/14046

quantimnot:


fox0430:
@quantimnot 

Thank you!

PR #635, opened 2020-05-27
Added half page down and half page up movements, that work with ctrl+…

tobimensch:
…d an ctrl+u.

This is one of my favorites of the basic vim movements.


fox0430:
@tobimensch 

Thank you very much !

PR #636, opened 2020-05-27
Implemented ctrl+u vim movement for insert mode.

tobimensch:
Another one of my favorite vim movements.

Here I used my own experimental code formatting style,
please feel free to change all or any part of it.

Oh, and I also added _ for navigating to the first non-blank character in the current line,
which is an alias for ^, that is hard to type on some keyboards.

And I added g_ which moves to the last non-blank character in the current line,
which is different from $ that moves to the last character whatever it may be.


fox0430:
@tobimensch 

Thanks a lot !

PR #638, opened 2020-05-28
Added vim D as alias for existing d$, new operation S (and an alias cc)

tobimensch:
Also added _ as alias for ^, which I apparently forgot to stage
for the last commit,
small reordering of code in normalmode.nim by using templates.

Please note, that D and S operations so far don't yank
anything, I noticed that existing operations such
as d$ and d0 don't do it yet either, so we have to add
this functionality for all of them.


fox0430:
@tobimensch 

Thank you !

tobimensch:
I have to thank.

Issue #646, opened 2020-05-29
Support for VSCode themes


tobimensch:
Reasoning:

A big userbase is needed to get a big number of themes from contributors.

Therefore it would be wise for moe to at least partially support
the theme format of one of the big editors/IDEs.

Because of that I'm currently working on vscode theme support for moe,
and will make a pull request for the partial support I have already implemented.

Why Vscode/VsCodium?

Their userbase is big and growing fast, there are tons of themes for it, and I happen to
use Vscode, so it was natural for me to want to have the same theme in moe and Vscode.

tobimensch:
I've implemented initial support for this feature in #648.

fox0430:
Completed

PR #648, opened 2020-05-31
Initial VSCode Theme support - Implement #646

tobimensch:
This was only tested against the Forest Focus theme for VSCode/VSCodium,
and was made to approximate the theme, not to imitate it perfectly,  (this is work in progress).
There are some rough corners.

It might work fine with other VSCode themes, but it's untested so far.
Help appreciated!

The vscode theme can be detected automatically, but support for this was
only tested on Linux. For this you simply set "vscode" as the theme in
the Standard section of the config.
You can set the baseTheme setting of the Theme section in the config
to a VSCode theme file and set the theme in the Standard section to "config",
this should work on all systems.

I created a macro named setEditorColor, that should make it relatively easy to
fix the code for more themes.
Here's one example:
```
    setEditorColor searchResult:
      foreground:
        adjust: ReadableVsBackground
      background:
        colorFromNode(jsonNode{"colors", "tab.activeBorder"})
```
As you can see "tab.activeBorder" is used for the background of search results,
which is a very creative mapping of colors from VSCode to moe, that randomly
works for the Forest Focus theme (and my taste), but will probably fail for other
themes. "adjust: ReadableVsBackground" ensures, that foreground colors are
readable compared to their background by doing some "magic" calculations and
adjusting the foreground color.
If we find some other VSCode theme properties that work better for a great number
of themes, they should be put first in the respective foreground/background lists,
as every statement there that doesn't return Color.default will be deemed a success,
and will thus be used to set that color.
Example:
```
    setEditorColor searchResult:
      foreground:
        colorFromNode(jsonNode{"colors", "some.betterForegroundForSearch"})
        adjust: ReadableVsBackground
      background:
        colorFromNode(jsonNode{"colors", "some.betterBackgroundForSearch"})
        colorFromNode(jsonNode{"colors", "tab.activeBorder"})
```
With these changes (but first we would have to know what that better property is!),
"tab.activeBorder" will simply be ignored, for all VSCode theme files, that contain
"some.betterBackgroundForSearch". And if "some.betterBackgroundForSearch" is
missing (like in Forest Focus) "tab.activeBorder" would be used as a fallback.

You can ask me anything.

### DIFF
--- a/src/moepkg/color.nim
+++ b/src/moepkg/color.nim
@@ -1,263 +1,399 @@
 import ncurses
+import strutils
+import tables
+import macros
 
-type Color* = enum
-  default             = -1
-  black               = 0
-  maroon              = 1
-  green               = 2
-  olive               = 3
-  navy                = 4
-  purple_1            = 5
-  teal                = 6
-  silver              = 7
-  gray                = 8
-  red                 = 9
-  lime                = 10
-  yellow              = 11
-  blue                = 12
-  fuchsia             = 13
-  aqua                = 14
-  white               = 15
-  gray0               = 16
-  navyBlue            = 17
-  darkBlue            = 18
-  blue3_1             = 19
-  blue3_2             = 20
-  blue1               = 21
-  darkGreen           = 22
-  deepSkyBlue4_1      = 23
-  deepSkyBlue4_2      = 24
-  deepSkyBlue4_3      = 25
-  dodgerBlue3_1       = 26
-  dodgerBlue3_2       = 27
-  green4              = 28
-  springGreen4        = 29
-  turquoise4          = 30
-  deepSkyBlue3_1      = 31
-  deepSkyBlue3_2      = 32
-  dodgerBlue1         = 33
-  green3_1            = 34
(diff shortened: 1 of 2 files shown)


fox0430:
@tobimensch 

Excellent work!
Thank you for your PR!

fox0430:
@tobimensch 

Can I write documentation?
You don't have to do a perfect job.

tobimensch:
@fox0430 
All help is much appreciated. :-)

fox0430:
@tobimensch 

I can not set theme...
This config is incorrect?

### ~/.config/moe/moerc.toml
```
[Standard]

theme = "vscode"
```

VS Code theme is Forest Focus.

tobimensch:
Yes, that should work.

What system are you running this one?

Do you have VSCode or VSCodium?

fox0430:
@tobimensch 

I'm using VSCode on Arch Linux.

tobimensch:
@fox0430 
I pushed another commit. Please try again, I set the wrong directory for VSCode.

fox0430:
It works! Thank you.

PR #652, opened 2020-05-31
Added icons to filer mode using unicode symbols.

tobimensch:
This is purely for the looks.


fox0430:
@tobimensch 

Nice. This is important ✨

Issue #653, opened 2020-05-31
Directories before files in filer


tobimensch:
I want to discuss this first, so this is the question I have:

Should directories be listed before files in filer mode?
What would speak against this change?
Or could it be an option?

fox0430:
@tobimensch 

It's a specification for reproducing netrw.
Not required.

tobimensch:
Do you know what the reasoning in netrw is for not separating dirs/files in the listing?
Would you be against changing the default?

fox0430:
@tobimensch 

I don't know...
It is unnecessary for practical use, so you can delete it.

PR #654, opened 2020-05-31
Keep directories before files. Discussed in #653

tobimensch:
As discussed in #653.

This always keeps the list of directories before the list of files.

The sorting still works, but it keeps dirs and files separated.

Hopefully I didn't misunderstand you, maybe you don't like
the idea.


fox0430:
@tobimensch 

No problem. Thank you!

PR #667, opened 2020-06-03
searching forward/reverse with # and *. WIP.

tobimensch:
this is only preliminary support. I think the searches aren't being added to the search history right now,
it doesn't bother me, though.

This also contains a fix, that should make the ctrl+a/ctrl+x thing more reliable.


tobimensch:
Please merge this, if possible, I marked it as WIP. But it should be stable.

fox0430:
@tobimensch 

Ok. Thank you.

PR #788, opened 2020-07-09
Disable the assertion in highlight.nim for performance

walkre-niboshi:


fox0430:
@walkre-niboshi 

Thank you.

PR #805, opened 2020-07-11
Information for installing latest sources from github with nimble

tobimensch:
Information for installing latest sources from github with nimble


fox0430:
@tobimensch 

Thank you!

PR #813, opened 2020-07-13
fix: visual and command mode messages

tjbell:
Adjusted the spelling of three functions in `visualmode.nim` and a few messages in `commandview.nim`.


fox0430:
@tjbell 

Thank you very much! That's very helpful.

PR #815, opened 2020-07-13
Recent file selection (#760)

### LINKED ISSUE #760: Recent files selection
This is something I often use in other editors, so I'd be disappointed if this was never getting added to moe.

Basically I want a filer like selection mode for the most recently opened files.

There is a freedesktop specification for Linux https://specifications.freedesktop.org/recent-file-spec/recent-file-spec-0.2.html , that hopefully works for most desktop environments. I suppose Windows and Mac also have something like it, so it should be supported, too. Personally I'm only using Linux though.

Unfortunately this specification seems to be out of date now.

On my system some recent files are found in:
.local/share/recently-used.xbel

But the spec says `The document should be stored in "~/.recently-used"`.

Also you'd have to investigate differences between how Gnome/KDE and lesser known desktops such as Enlightenment handle it. It might be a bit of a mess, but the feature should be worth the trouble.

I would like to open this selection mode with a command like:
```:recent```

fox0430:
Close #760

### DIFF
--- a/documents/howtouse.md
+++ b/documents/howtouse.md
@@ -201,3 +201,5 @@ Check command bar
 ```putConfigFile``` - Put a sample configuration file in ~/.config/moe
 
 ```run``` - Quick run
+
+```recent``` - Open recent file selection mode (Only supported on Linux)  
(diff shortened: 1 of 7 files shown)


tobimensch:
It works! I thank you.

Issue #830, opened 2020-07-20
Cursor bug


unihernandez22:
The cursor changes when opening moe, but no when closing. So the terminal cursor changes.

tobimensch:
What terminal are you using?

Can you give detailed steps of what you did and the outcomes?

fox0430:
@unihernandez22 

Please try set ```Standard.defaultCursor``` settings.
Do not support auto restore cursor shape yet.

unihernandez22:
I had the problem on Peppermint Linux, Sakura is the terminal emulator. I think the best solution is don't change the cursor, like vim.

unihernandez22:
> Please try ```Standard.defaultCursor``` setting.

Ahhh, ok.

walkre-niboshi:
I think the better solution is to provide an option to disable the change of the cursor, like neovim(https://github.com/neovim/neovim/wiki/FAQ#how-to-change-cursor-shape-in-the-terminal).

fox0430:
I will add the setting to disable the change of the cursor shape.

PR #832, opened 2020-07-20
Add setting to disable the change of the cursor shape

fox0430:
Link #830


unihernandez22:
Why `disableChangeCurosr` instead of `disableChangeCursor` ?

fox0430:
@unihernandez22 

This is a typo.
Thank you for teaching me.

PR #836, opened 2020-07-21
Fix #831

### LINKED ISSUE #831: Unhandled exception when opening /mnt/c/Users/ in WSL
I opened `/mnt/c/Users/` in WSL:
```
walkre@DESKTOP-07UPQ8F:moe [develop]$ ./moe /mnt/c/Users/
/mnt/c/Users/walkre/code/projects/moe/src/moe.nim(69) moe
/mnt/c/Users/walkre/code/projects/moe/src/moe.nim(60) main
/mnt/c/Users/walkre/code/projects/moe/src/moepkg/filermode.nim(459) filerMode
/mnt/c/Users/walkre/code/projects/moe/src/moepkg/filermode.nim(133) updateDirList
/mnt/c/Users/walkre/code/projects/moe/src/moepkg/filermode.nim(106) refreshDirList
/mnt/c/Users/walkre/code/projects/moe/src/moepkg/fileutils.nim(5) normalizePath
/home/walkre/.choosenim/toolchains/nim-1.2.4/lib/system/fatal.nim(49) sysFatal
Error: unhandled exception: index out of bounds, the container is empty [IndexError]
```

fox0430:
Close #831

### DIFF
--- a/src/moepkg/filermode.nim
+++ b/src/moepkg/filermode.nim
@@ -80,10 +80,13 @@ proc sortDirList(dirList: seq[PathInfo], sortBy: Sort): seq[PathInfo] =
     result.add dirList.sortedByIt(it.lastWriteTime)
 
 proc refreshDirList(sortBy: Sort): seq[PathInfo] =
-  var dirList  : seq[PathInfo]
-  var fileList : seq[PathInfo]
-  var item     : PathInfo
+  var
+    dirList  : seq[PathInfo]
+    fileList : seq[PathInfo]
+
   for list in walkDir("./"):
+    var item: PathInfo
+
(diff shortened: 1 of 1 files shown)


walkre-niboshi:
And this is slightly another matter, in my opinion, even if it fails to expand the symbolic link I would like to see at least the existence of it.

walkre-niboshi:
The existence of symbolic links that fail to be expanded is now certainly displayed in filer mode. However, it shows up as an empty string. I think this is not very desirable behavior, but more should be discussed in another issue or pull request.

Anyway, opening /mnt/c/Users/ in my environment now no longer throws an exception. That is, I think this issue has probably been resolved.

Thank you for your work!

PR #894, opened 2020-07-29
VSCode theme: Fix git branch name color in the status bar (#893)

### LINKED ISSUE #893: Regression with vscode theme:  branch name black and not visible in status line.
@fox0430 I think you added a new theming/color setting. This has affected the vscode theme feature, too.

It used to work in the last version, but in current head, the branch name in the statusbar/statusline now is totally black, and the background as well, so it has become invisible/blacked out.

fox0430:
Close #893


tobimensch:
Okay, I tested this, and it works definitely better now.

PR #897, opened 2020-07-30
Fix lastWriteTime of symbolic links

walkre-niboshi:
Use `getLastModificationTime(list.path)` as lastWriteTime of symbolic links instead of `getLastModificationTime(getCurrentDir())`.


fox0430:
@walkre-niboshi 

Thanks

PR #903, opened 2020-07-30
QuickRun: Fix #899

### LINKED ISSUE #899: QuickRun not working for simple nim test
1. Open moe
2. Create these lines
```
proc test() =
  echo "test"
test()
```
3. Save with ```:w test.nim```
4. ```:run```

The QuickRun pane/window opens on the right with no output shown. Expected output is "test".

When I close moe again and open the saved test.nim file,
and do a ```:run``` again, it finally works.

fox0430:
Close #899


tobimensch:
Works, thank you.

PR #906, opened 2020-07-31
Ex mode: Fix sp command (#904)

### LINKED ISSUE #904: sp command is in ex command list, yet it says "not an editor command"
When you ```:s``` + tab, sp is listed as one of the commands,
but when you try to use it, you get this message instead:
```Error: Not an editor command: sp```

fox0430:
Close #904


tobimensch:
Fixed! Thank you

PR #917, opened 2020-08-02
Quick Run: Showing the output even if QuickRun faile (#916)

### LINKED ISSUE #916: Showing the output even if QuickRun failed
Currently, moe doesn't show the output if QuickRun failed.
Showing the output even if QuickRun failed, we will be able to do trial and error more speedily. This seems quite convenient for me.

fox0430:
Close #916


walkre-niboshi:
It seems to work well. Thank you!

PR #940, opened 2020-08-07
Fix #936

### LINKED ISSUE #936: Automatic backups can crash if the file path contains non-US-ASCII characters

walkre-niboshi:
Fix some wrong procs in unicodeext.nim and add tests for them.

Closes #936


fox0430:
@walkre-niboshi 

Thanks

PR #989, opened 2020-09-05
Add general-purpose autocomplete

### LINKED ISSUE #927: Simple auto complete
Auto completes a word in the buffer. This can be used regardless of language.

walkre-niboshi:
Closes #927.


walkre-niboshi:
The below list is features that I think should be implemented, in the future.
- Add a setting for the key used to move up and down in the suggestion window.
- Make the suggestion list asynchronously.
- Make it possible to change the suggestion window color.
- Make it possible to set the max height/width of the suggestion window.
- Make it possible to set the max length of the suggestions.
- Make it possible to select the algorithm used to create the suggestion list.

fox0430:
@walkre-niboshi 

Thank you so much. This feature is absolutely cool!

fox0430 (approved):
I think it's almost good.

PR #1129, opened 2020-10-31
Change setup-nim-action version

jiro4989:
- Using setup-nim-action `v1`
- Add paths-ignore
- Add `skip` job
  - `before` job will be skipped when commit message has `[skip ci]`. But GitHub Actions raises error when all jobs were canceled. So `skip` job needs. `skip` job always success.

PR #1130, opened 2020-10-31
Fix before job #1129

jiro4989:


jiro4989:
Ah, wait

jiro4989:
CI may fail again when commit message has `"` (double quote).

jiro4989:
```bash
echo -e "foobar "

"
```

jiro4989:
umm

jiro4989:
Removed printing commit message.

fox0430:
@jiro4989 

Ok. Thanks!

PR #1285, opened 2021-03-08
fix a typo in previous PR (#1257)

leetschau:
I remember a PR should be auto-refreshed when there're new commits in the downstream repository.
However, PR #1257 doesn't change accordingly after I push a fixing commit.
So I create a new one.
Please tell me if there're better ways in this situation.
Thanks.


fox0430:
@leetschau 

It’s okay. No problem.

PR #1396, opened 2021-07-13
Enhance autoindent

### LINKED ISSUE #874: Enhance Autoindent
Auto indent the next line when:
- The line is `var`, `let` or `const`
- The line finishes with `or`, `and` or `object`
- The instruction isn't complete (eg: Parenthesis aren't closed)

And this: (`<>` represents the indentation)
```rust
fn main() {}
```
(Cursor is between `{` and `}`)
`<Return>`
```rust
fn main() {
<>
}
```

or

```nim
echo someFunc()
```
(Cursor is between `(` and `)`)
`<Return>`
```nim
echo someFunc(
<>
)
```

See https://github.com/zah/nim.vim

fox0430:
Enhance auto-indent for some languages (Nim, C, C++, C#, Java, JS, Yaml).

Close #874

### DIFF
--- a/src/moepkg/editor.nim
+++ b/src/moepkg/editor.nim
@@ -187,8 +187,72 @@ proc deleteBeforeCursorToFirstNonBlank*(bufStatus: var BufferStatus,
   for _ in firstNonBlank..max(0, windowNode.currentColumn-1):
     currentLineDeleteCharacterBeforeCursor(bufStatus, windowNode, false)
 
+proc isWhiteSpaceLine(line: seq[Rune]): bool =
+  result = true
+  for r in line:
+    if not isWhiteSpace(r): return false
+
+proc deleteAllCharInLine(line: var seq[Rune]) =
+  for i in 0 ..< line.len: line.delete(0)
+
+proc basicNewLine(bufStatus: var BufferStatus,
+                  windowNode: WindowNode,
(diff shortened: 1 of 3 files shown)


src/moepkg/editor.nim
```diff
@@ -279,19 +251,259 @@ proc keyEnter*(bufStatus: var BufferStatus,
...
+    line = bufStatus.buffer[currentLine]
+
+  if line.len > 0:
+    # Auto indent if the current line are "var", "let", "const".
+    # And, if finish the current line with ':', "object"
+    if (currentColumn == line.len) and
+       (line.splitWhitespace == @[ru "var"] or
+       line.splitWhitespace == @[ru "let"] or
+       line.splitWhitespace == @[ru "const"] or
+       (line.len > 6 and line[line.len - 6 .. ^1] == ru "object") or
```
uninhm, line 322: Why not:
`line.splitWhitespace[^1] == (ru "object") or`

fox0430: You are right. This should be ```line.splitWhitespace[^1] == (ru "object") or```

uninhm: Solved #1422

fox0430: Thanks!

PR #1419, opened 2021-08-16
Add '=' case to autoindent

uninhm:
It's a very little change but I couldn't wait for it


fox0430:
@uninhm 

Nice! Thank you!

PR #1420, opened 2021-08-16
Fix crash on '\' press

uninhm:


fox0430:
@uninhm 

Thank you very much!

PR #1478, opened 2022-04-14
Fix a few typos

0branch:


fox0430:
@0branch 

Thank you for PR! It was very helpful.

PR #1517, opened 2022-11-16
[Enhancement] Configure Dependabot

### LINKED ISSUE #1516: Suggestion:  Dependabot
@fox0430

At the moment, I am working on the obsolete conversions and redundant imports.  Thereby, I noticed that some of the Actions the CI requires have received major updates.  Some of the current versions are not supported anymore, thus, the CI might break if not updated.

I would like to suggest to submit a separate Pull Request with the version updates such that the CI will continue to work.

In the long term, would Dependabot be useful in order to keep all dependencies up-to-date?  Once configured, it will submit Pull Requests with the respective updates.  At option, you can write it an email (many mailing programs name this "reply to list") in order to instruct it to merge the update.  I previously worked with Dependabot in some other repositories, especially to track GitHub Actions, so I can prepare an initial setup, if you would like to try out this feature.  I therefore would need to know which settings are preferred, especially:

* the frequency of checks for updates (I recommend daily in order to do not miss important updates.)
* the time when to check for updates (I highly recommend specifying a time; otherwise, the time will be random.  This can be either some UTC time or a time relative to a given time zone.)

Shall I prepare an initial setup for Dependabot?

kevinmatthes:
This Pull Request introduces an initial configuration for Dependabot and thereby fixes #1516.  The initial configuration is as follows:

* Dependabot will check for updates each day, 00.00 UTC.
* The commits authored by Dependabot will include the new version numbers.  This is useful for documentation and future reference.
* Dependabot will assign @fox0430 on each Pull Request.  This allows for email replies with instructions for Dependabot.

In order to instruct Dependabot to merge something, you need to end a Markdown paragraph with a mention of Dependabot and the appropriate instruction.  The three most important are self-explanatory:

* `@dependabot squash and merge`
* `@dependabot rebase`
* `@dependabot merge`

Each Pull Request always contains a quick reference of all available commands, just expand the respective section in the initial comment.

Important:  some email-providers seem to reject subsequent messages with the same content.  Thus, the Dependabot instructions may need to differ; just append some dummy text to ensure the email to be sent.  Example:

```
Hello!

@dependabot merge

Thanks!
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+      time: '00:00'
+    assignees:
+      - fox0430
+    commit-message:
+      include: scope


kevinmatthes:
By default, it will raise at most 5 Pull Requests against the default branch of this repository.  Both can be altered.  Further options are:

* milestones
* commit prefix (default:  `chore`)
* certain labels (Dependabot will create its own label `dependencies` together with one with the name of the ecosystem, i.e. `github_actions`)
* additional ecosystems (Git sub-modules, Python packages, JavaScript packages, Docker, ...)

There is a whole section on [Dependabot](https://docs.github.com/en/code-security/dependabot) in the official documentation containing more information, if required.

fox0430:
@kevinmatthes 

Thanks!

kevinmatthes:
@fox0430

Dependabot sometimes needs a little time to start working, a delay of about one or two days for a Git sub-module is not unusual.  The sole exception is when its configuration file was edited; then, it will immediately check for updates.

By default, the creation of `.github/dependabot.yml` suffices to activate Dependabot for a repository.  The reason why it does not react might be that the file was not created by a maintainer but by a contributor instead.  You can make some dummy changes to the file (e.g. add a blank line between line 2 and 3) and commit it to default branch to cause Dependabot to check for validity.  This should suffice.

fox0430:
@kevinmatthes 

The dependabot config seems to be broken.

Ref: https://github.com/fox0430/moe/runs/9535932927

fox0430:
@kevinmatthes 

Fixed by #1521 and active now.

kevinmatthes:
@fox0430

I apologise for having caused these problems.  I did not mind that `commit-message.include` requires `commit-message.prefix`.  I am sorry.

fox0430:
@kevinmatthes 

No problem. Thank you for your contribution!

Issue #1514, opened 2022-11-15
Build: Common Nimble Hints


kevinmatthes:
When working on #1513, I had a look at the output of the CI.  The step "unit test" of the job "build ubuntu-latest" (`test.yaml`) generates various hints on style issues Nimble finds.

Here is an overview about common categories:

* `[ConvFromXtoItselfNotNeeded]`
* `[DuplicateModuleImport]`
* `[Name]`
* `[XCannotRaiseY]`
* `[XDeclaredButNotUsed]`

The unit test generates more than 8000 lines of output, so I picked the most frequently appearing categories, for the beginning.  Some points, like inconsistent spacing, are rather simple to fix and were only rarely reported in the CI's output.

The naming convention hints mostly concern the colours, but also other symbols in the code, such as the options for compilers (`ClangOptions` → `clangOptions`, for example).  Since most of the hints Nimble generates are related to the naming conventions, I would like to suggest to defer the adjustment to dedicated enhancements of the concerning components and major refactorings, respectively, for example #1483, as the reason for the choice of a symbol's name for sure is more important than the respective naming convention being fulfilled in mostly any case.  However, an adjustment would be a meaningful change in the long term as the compliance to the common conventions of the language is a sane goal, in my opinion.

The obsolete conversions and redundant imports are rather easy to fix, I assume that especially the removal of the casts would lead to some performance benefits.  Again, often the colour management was affected.

Regarding the declared but unused symbols as well as the information that some procedures cannot raise certain exceptions I am unsure which of those points might belong to a long-term feature introduction and / or maintenance.  Here, it should be discussed which changes can be made without breaking pending Pull Requests.  Mostly, `b2Mask`, `b3Mask`, and `b4Mask` were flagged declared but never used; `parse`cannot raise `Defect`.

I would like to discuss how these points should be prioritised and which issues should be worked on first.

fox0430:
@kevinmatthes 

Thank you for the report.
I think it's good to start with easy tasks. For example, obsolete conversions and redundant imports.
And, It may be a good idea to remove the hints displayed by `nimble build --verbose` first. There don't seem to be many of these.
Of course, If you are interested in something, you can start with that.

kevinmatthes:
@fox0430

Thank you for your answer.

I will begin to remove the obsolete conversions and redundant imports.  As soon as I finished these adjustments, I will submit a new Pull Request.

kevinmatthes:
[...]

PR #1518, opened 2022-11-16
[Bugfix] Fix Obsolete Conversions and Most Redundant Imports (#3)

kevinmatthes:
> This Pull Request is related to #1514.

As discussed in #1514, this Pull Request fixes all obsolete conversions and most of the redundant imports.  Exactly 50 redundant imports are not obviously resolvable as they are introduced due to the inclusion (`include`) of the respective source files.  Resolving them requires major changes in the design, thus, I focused on those redundancies which were obvious.

Despite 1 left occurrence, I also resolved all `[UnusedImport]`s.  The sole one left is the import of `strformat` in `highlight.nim` which is only found during the build.  Thus, the source files including / importing `highlight.nim` need to import `strformat` on demand.  This should be subject to another Pull Request as it can be combined with further changes to the design.

I only removed the imports but the list brackets are still present, this is, for instance,

```nim
import std/[unittest]
```

@fox0430, shall I tidy this up or is this style okay?  I decided to leave it that way in case further packages of the same namespace need to be included.

### DIFF
--- a/src/moepkg/editorview.nim
+++ b/src/moepkg/editorview.nim
@@ -223,8 +223,8 @@ proc writeCurrentLine(win: var Window,
     block:
       let
         fg = colors[0]
-        bg = if colors[1] == theme.EditorColor.editorBg:
-               theme.EditorColor.currentLineBg
+        bg = if colors[1] == theme.editorBg:
+               theme.currentLineBg
              else:
                colors[1]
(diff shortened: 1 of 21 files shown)


fox0430:
@kevinmatthes 

Thanks! no problem with this style.

PR #1522, opened 2022-11-17
[Bugfix] Adjust C++ Source File Extensions

kevinmatthes:
The C++ mode did not recognise source files with the extensions `*.hpp` and `*.cc`, at the moment.  The reason therefore is that the extensions in `src/moepkg/highlight.nim`, line 301, were not prefixed with `.`.  This Pull Request fixes this issue.  Only the C++ mode was affected, all other modes have their extensions prefixed appropriately.

By the way, there are also various other common extensions for the already configured languages.  Shall I create a list of those?

### DIFF
--- a/src/moepkg/highlight.nim
+++ b/src/moepkg/highlight.nim
@@ -298,7 +298,7 @@ proc detectLanguage*(filename: string): SourceLanguage =
     return SourceLanguage.langNim
   of ".c", ".h":
     return SourceLanguage.langC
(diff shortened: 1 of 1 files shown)


fox0430:
@kevinmatthes 

Thank you!

fox0430:
@kevinmatthes 

It would be helpful if you could make the extensions list.

kevinmatthes:
@fox0430

I will create a new issue therefore.

PR #1525, opened 2022-11-17
[Enhancement] Add Support for Haskell (#4)

### LINKED ISSUE #1116: Haskell highlighting

kevinmatthes:
This Pull Request adds support for Haskell and thereby fixes #1116.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This project's goals are easily customizable, high productivity, user friendly, 
 
 - Infinite undo/redo
 
-- Syntax highlighting (Nim, C, C++, C#, Java, Yaml, JavaScript, Python)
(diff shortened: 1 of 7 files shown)


kevinmatthes:
As all symbols of the `Prelude` can be both explicitly imported and redefined, I did not add them to the list of keywords.  The keywords originate from https://github.com/vigna/ne/blob/master/syntax/haskell.jsf.  Haskell preprocessing instructions are supported!

kevinmatthes:
`ne` is an editor written in C and licensed GPL-3.0.  Thus, reusing their collection of keywords should not be a problem as furthermore, they use an automaton description language while this suggestion is written in pure Nim (inspired by the already implemented JavaScript highlighting logic).

fox0430:
@kevinmatthes 

Looks good! Great works!

Issue #161, opened 2019-03-03
Syntax highlighting enhancement


fox0430:
Add keywords to highlighting

walkre-niboshi:
@fox0430  That should be identifiers instead of keywords, shouldn't it?

fox0430:
Yes

walkre-niboshi:
@fox0430 OK. I will attempt to highlight only type names.

kevinmatthes:
@fox0430 I noticed a small bug related to this issue:  Python comments are not highlighted as such.  I will have a look at that in the next days.

kevinmatthes:
@fox0430

I am about to have a look at the comment parsing logic since there are several redundancies which can be fixed quite elegantly.  Thereby, the new logic will be stored as a set of procedures in a new module, `src/moepkg/syntax/comments.nim` as these are going to be quite a lot functions.

In parallel, I would also turn the character types (decimal characters, hexadecimal ones, ...) into central public constants in order to reduce those redundancies, as well.  Concerning this step, I have got a question:  shall I turn this into a dedicated module, as well, or should they just be added to `src/moepkg/syntax/highlite.nim`?

fox0430:
@kevinmatthes 

> In parallel, I would also turn the character types (decimal characters, hexadecimal ones, ...) into central public constants in order to reduce those redundancies, as well. Concerning this step, I have got a question: shall I turn this into a dedicated module, as well, or should they just be added to `src/moepkg/syntax/highlite.nim`?

If it's large, a dedicated module may be better, and if it's small,  `src/moepkg/syntax/highlite.nim` may be better.
I think it's fine as you like.

kevinmatthes:
@fox0430

Would it be okay if documentation comments would be highlighted in a different colour in order to distinct them from ordinary comments?

fox0430:
@kevinmatthes 

Of course. It's a good idea!

kevinmatthes:
@fox0430

The refactoring of the Haskell highlighting logic takes longer than I originally expected.  At the moment, I am working on turning the comment highlighting logic into a more general lexing logic.

PR #1527, opened 2022-11-18
[Bugfix] Adjust Python Syntax Highlighting (#7)

kevinmatthes:
As discussed in #161, I adjusted the Python syntax highlighting regarding the comments.

- Line comments are now recognised as such.
- Python's nested comments are unbound literal string constants, no changes were necessary.
- The Python shebang is an important line for UNIX systems.  I turned it into a preprocessor directive for highlighting.


fox0430:
@kevinmatthes 

Thank you!

PR #1533, opened 2022-11-19
[Enhancement] Maintain Comment Parsing Logic (#10)

kevinmatthes:
This PR maintains the comment parsing logic.  The logic is now outsourced into a dedicated module to enhance both its re-usability and maintainability.  The behaviour is controlled by flags which can be set per language.

As discussed in #161, documentation comments are now rendered in a different colour as they are different from ordinary comments.  I set it to the colour of string literals, initially, but changing this is very simple.

I wired up the new logic for both Python and Nim.

### DIFF
--- a/src/moepkg/syntax/comments.nim
+++ b/src/moepkg/syntax/comments.nim
@@ -1,0 +1,233 @@
+#[###################### GNU General Public License 3.0 ######################]#
+#                                                                              #
+#  Copyright (C) 2017─2022 fox0430                                             #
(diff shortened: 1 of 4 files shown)


kevinmatthes:
@fox0430

I also added the GPL-3.0 license header to the newly created source file `comments.nim`.  This is a mandatory requirement by the LICENSE, ll. 629.

fox0430:
@kevinmatthes 

Great job! Thank you!

PR #1543, opened 2022-11-22
[Enhancement] Refactor Comment Lexing Logic (#13)

kevinmatthes:
This PR restructures the comment lexing logic.

At the moment, the logic I introduced by #1533 focuses only on processing comments.  But with some extensions, it can also lex other syntax elements, such as Rust's annotations.  Therefore, I split the recent implementation into dedicated modules whose logic is controlled by the flags which describe a language.

The long term plan is to create a lexing function for each character type with the respective behaviour only differing due to the flags.  This will simplify the creation of new highlighting schemes.


kevinmatthes:
@fox0430

The latest commit adds part one of the Haskell comment analysis logic refactoring:  the difference between simple and documentation line comments is now respected.

kevinmatthes:
@fox0430

The refactoring of the comment lexing logic in general has the following reasons:

- I did not mind the difference between parsing and lexing.  A lexer produces tokens, a parser processes them.  And here, tokens are created, thus the function renaming.  I am sorry, it was my mistake.
- My previous design was hardly maintainable and, due to the issue regarding large files, I assumed that one module would not suffice for all lexing procedures.  I know a related issue of the Zee editor which truncated too large files on save by accident; this is why I separated the logic into further but smaller files in order to prevent such a truncation in Moe's own source code.

fox0430:
@kevinmatthes 

Okay! Thank you for the PR!

PR #1544, opened 2022-11-24
[Added] CHANGELOG:  Setup (#15)

### LINKED ISSUE #1542: Discussion:  CHANGELOG?
@fox0430

Should we add a CHANGELOG to Moe?  There are also utilities which assemble the CHANGELOG on release and write the GitHub release notes -- in addition to GitHub's built-in release note generation.

kevinmatthes:
This PR fixes #1542.

The style is based on Keep a Changelog, a modern and intuitive community standard many projects apply.  The format is ReStructuredText which is also the native format of Nim documentation comments.

As many projects add a reference to the PRs which introduced the respective changes, I did so, as well.  I derived the entries from all PR titles since the latest stable version was release, except for the mandatory version number increment which I considered obvious; thus, it is not listed.

The overall style is an initial, concrete suggestion for further discussion.  In case something should not be optimal, I can push appropriate changes.  If everything is fine for you, @fox0430, you can directly merge it.


kevinmatthes:
@fox0430

For this CHANGELOG style, there are free and open source management tools available in order to simplify the maintenance.  Shall I create a discussion issue with further details?

fox0430:
@kevinmatthes 

Thank you for the PR. It's very helpful.

> 
> For this CHANGELOG style, there are free and open source management tools available in order to simplify the maintenance. Shall I create a discussion issue with further details?

Yes. please.

kevinmatthes:
@fox0430

The latest commit brings the Scriv support together with the workflow, as discussed in #1545.  The commit only introduces support for Scriv, I did not configure bump2version, yet.

Shall I add it, as well?  If so, I would like to suggest to do so in another Pull Request.

kevinmatthes:
@fox0430

I now found a way to combine both approaches in one file.  I will submit it in a few minutes.

kevinmatthes:
@fox0430

You can now assemble the CHANGELOG *and* increment all versions at once.  Just go to the Actions tab, select this workflow, and click "Run workflow".  Then, select the release type you want (major, minor, patch) as well as the branch to commit to and click the green button "Run workflow".

Please do not forget to set the release date in the CITATION.cff appropriately; I did not found a way to automate this, yet...  As the workflow saves much development time, this should not be a major problem, however.

kevinmatthes:
@fox0430

That's it.  Finally.  Seriously.

- Update the CFF release date on release by a GitHub Action.
- Assemble the CHANGELOG on release by the same GitHub Action.
- Increment all hard-coded version numbers by the same GitHub Action.

moe.nimble
```diff
@@ -1,6 +1,6 @@
 # Package
 
-version       = "0.4.0"
+version       = "0.3.0"
```
fox0430, line 3: @kevinmatthes 

v0.3.0 is already released. Should we still use it?

kevinmatthes:
@fox0430

Is there something about this PR I need to change?  Basically, from my side, all important changes are done, so you can review and merge it, if you agree with the changes.

fox0430:
@kevinmatthes 

As I pointed out in the review, why are you using v3.0.0 in moe.nimble?

kevinmatthes:
@fox0430

I am sorry, I did found any review; neither here on GitHub nor in the GitHub notifications.

The version decreasing is required in order to combine all steps.  The version number increment requires a homogeneous version number in all tracked files.  By now, the latest version `v0.3.0`, so I made this the current version.  On release, the version will be incremented back to `v0.4.0`.

kevinmatthes:
@fox0430

I just checked the GitHub notifications as well as this PR again; there is no review visible for me.  I did not receive any messages for this PR in the last days and at the top of this PR, the "Reviewers" section in the right side bar tells me "No reviews".

Would you please re-review this PR such that I can see your notes?

fox0430:
@kevinmatthes 

Okay, I understand it. Thank you.

And I have some questions. Honestly, I don't understand these mechanisms completely yet.

1. We add fragments for scriv in every PR. Is this correct?
2. We run the workflow for updating the CHANGELOG in a PR for release. Is this correct?
3. How version is in develop branch after releasing the next version (v0.4.0)? I expect will be v0.5.0. 

Sorry if you already explained.

kevinmatthes:
@fox0430

1. Yes, this is correct.
2. Basically:  yes, the workflow will update the CHANGELOG for release.  But:  doing so in a new PR is possible but not mandatory.  You select the target branch to commit to; this can be *any* existing branch.  In case of a PR, you will need to create a new branch, at first, commit to it, and raise a PR from it.  A PR allows for further adjustments, for instance adding the new version to a bug report GitHub Issue Form.
3. The last release of Moe was tagged with version 0.3.0.  This is what I consider the current version.  When you run the workflow, you will need to select the release type:  patch, minor, and major.  A patch release will increment all version numbers to 0.3.1, a minor release to 0.4.0, and a major release to 1.0.0.

Does this answer your questions?  I can also give some examples for these points to illustrate them.

kevinmatthes:
@fox0430, did you already click the green "Submit review" button in the top right corner of the changed files view?  Maybe this could be the reason?  I once had the same problem with GitLab -- I simply forgot to submit my review.

fox0430:
@kevinmatthes 

Thank you 1 and 2 are ok.
but I still don't know about 3.  Will the develop branch version after the v0.4.0 release be v0.4.0? Is it possible to bump to v0.5.0 in moe.nimble without problems?

moe.nimble
kevinmatthes: Yes, we need to do so.  That is the somehow confusing thing about bump2version:  when it returns, the version is incremented.  bump2version assumes the version to be updated ***right before*** the new release is published instead of directly after publishing it.

kevinmatthes:
@fox0430

> Will the develop branch version after the v0.4.0 release be v0.4.0?

Yes, that is correct.

> Is it possible to bump to v0.5.0 in moe.nimble without problems?

Technically, yes, it is possible.  In order to have a development version, some projects prefer to attach something like `-dev` to the version number, e.g. `v0.4.0-dev`.  Even bump2version itself does so; https://github.com/c4urself/bump2version.  Would this be the desired behaviour?

moe.nimble
kevinmatthes: @fox0430

I will create an example.

kevinmatthes: @fox0430

This is the example:  https://github.com/kevinmatthes/git-aliases-f18/pull/95.

The repository uses the very same workflow as I suggest here; the changes I submit with this PR are copied from there (which is no problem since it the repository is one of my own repositories).

I released the software in that repository with the version v0.1.3.  Let us assume that after some work, I can now release v0.2.0.  Please look at the changed files; https://github.com/kevinmatthes/git-aliases-f18/pull/95/files.  The whole time I worked on the changes for the new version, I still used the old version number.  I only increment it on release.  This requires that all tracked files have the same version number hard-coded somewhere.

This is why I needed to downgrade the version in `moe.nimble`:  because it will be re-upgraded on release.

(The example is also a preview of how a release with the suggested workflow will look like.  Please ignore the CI in the example; it cannot fetch Valgrind for memory integrity check and thus takes that long time before it will, probably, fail.)

fox0430: @kevinmatthes 

Thank you for the example. I understand it.

fox0430:
@kevinmatthes 

Okay. I understand it. Thank you for the answer.
No problem with this behavior. Probably my questions have been resolved and I will merge this PR soon.

Thank you for your contribution.

PR #1552, opened 2022-12-05
[Added] Workflow for Fresh Scriv Fragments (#21)

### LINKED ISSUE #1550: [Question] Scriv Fragment Creation Workflow?
@fox0430

Since the software requirements for the release workflow, bump2version and Scriv, are kept truly virtual, there is no need to install them.  But in order to create a Scriv CHANGELOG fragment for a PR, its templates might be useful.

Scriv assumes a certain structure for a fragment.  In order to not violate this structure, Scriv can create you a fresh fragment with the expected categories as comments.

Shall I create a workflow to generate fresh Scriv fragments with explanatory comments and the expected categories predefined in order to make it easier to add the corresponding fragment to a PR?

kevinmatthes:
This PR fixes #1550.

How it works:  go to the Actions tab, select the new workflow "Scriv" and use the manual dispatch.  Select the target branch to commit the new fragment to and run the workflow.

### DIFF
--- a/.github/workflows/scriv.yaml
+++ b/.github/workflows/scriv.yaml
@@ -1,0 +1,58 @@
+######################## GNU General Public License 3.0 ########################
+##                                                                            ##
+## Copyright (C) 2017─2022 fox0430                                            ##
+##                                                                            ##
+## This program is free software: you can redistribute it and/or modify       ##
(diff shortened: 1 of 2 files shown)


fox0430 (approved):
@kevinmatthes 

Thank you

PR #1596, opened 2023-01-22
[Enhancement] Adjust GitHub Action Workflows; Fix #1595 (#23)

### LINKED ISSUE #1595: [Bug Report] Trouble with Scriv Fragment Creation
@fox0430

The latest versions of Scriv require Python 3.11 to work.  Version 1.0.0 still works fine, though.  I am working on a patch for the GitHub workflows.

kevinmatthes:
This PR fixes #1595.

Furthermore, the Python 3 software requirements are now truly virtual and do not need to be specified in the repository, anymore.


kevinmatthes:
@fox0430

I made the GitHub Action workflows requiring Scriv mostly into dedicated Actions on their own.  If something should change regarding Scriv and / or bump2version, Dependabot will suggest you the updates to the Actions.  For support, please submit me an issue in the respective repositories:

* https://github.com/kevinmatthes/create-scriv-fragment
* https://github.com/kevinmatthes/release-bump2version-scriv

fox0430:
@kevinmatthes 

Nice work. Thank you.

Although not directly related to this issue, I noticed that reference numbers ( #23 in this PR ) in your PR title (commit message) were linked to the irrelevant issue or PR after merging. Is it possible to fix it?

kevinmatthes:
@fox0430

Yes, I can fix it, if you wish.

I prefer squashing my changes down to single commit before submitting the PR; therefore, I create an auxiliary branch which collects the commits to be squashed into the branch for PR submission.  GitHub automatically annotates the resulting commit with the number of the corresponding PR, and I did not change the default commit title since I consider this a useful additional information.  I assume that I should rather remove it in future PRs, as well, right?

fox0430:
@kevinmatthes 

Yes. remove it please. Thank you.

kevinmatthes:
@fox0430

Done.  This also updates one of the newly introduced Actions I wrote.

.github/workflows/release.yaml
```diff
@@ -41,37 +41,15 @@ jobs:
...
-            's/^date-released: ....-..-../date-released: '`date +%F`'/g' \
-            CITATION.cff >> tmp
-          mv tmp CITATION.cff
-
-      - name: scriv
-        run: scriv collect
-
-      - name: commit
-        run: git commit -am 'Increment ${{ inputs.release }} version'
+      - name: release-bump2version-scriv
```
kevinmatthes, line 49: I do not know why but the changes to this file will crash the editor reproducibly; try `moe .github/workflows/release.yaml` with this version of the file and the latest compilation of Moe from the main repository.

kevinmatthes: I assume that (here) line 52 is a very strange edge case of what YAML allows.  When developing the Action, I remember that GitHub's syntax highlighting always complained about the style.  It recommended me to put `${{ inputs.release }}` into quotes as something like a string would have been expected.  The Action itself worked, on the other hand.  So, I guess that the style is just somehow sloppy...

By the way, GitHub does not complain about the style, anymore; I assume they fixed that bug in the meantime.

fox0430: > I do not know why but the changes to this file will crash the editor reproducibly; try `moe .github/workflows/release.yaml` with this version of the file and the latest compilation of Moe from the main repository.

It seems to crash due to the single quote in line 42. So it's probably caused by the syntax highlighting in yaml going wrong. This is the bug and we should fix it in another PR.

fox0430: @kevinmatthes 

It seems to break syntax highlighting on neovim too from line 42... Is syntax highlighting no problem in the editor you are using?

kevinmatthes: @fox0430

I mostly edit files with Moe, now.  If I am making a somehow urgent change like this and happen to encounter a bug like the one described, I switch to another editor like `ne` (https://github.com/vigna/ne) or `mcedit` (part of the Midnight Commander, `mc`).  Both have YAML syntax highlighting and do not break at all.

To upload changes to GitHub, I copy the file contents from my local machine to the web editor.  As described, it only complained in an earlier version that quotes would have been missing but nothing more happened.

kevinmatthes: @fox0430

GitHub parses syntax with Linguist (https://github.com/github/linguist), `ne` uses just like `mcedit` an automaton description; https://github.com/vigna/ne/blob/master/syntax/yaml.jsf and `cat /usr/share/mc/syntax/yaml.syntax` with `mc` installed.

Maybe Moe's YAML highlighting fails due to the different handling of the three context types? An option might be to refactor it inspired by how `mcedit` does the highlighting.

fox0430: @kevinmatthes 

Thank you. I'll solve this problem in other PRs. First, as a temporary workaround, prevents crashes by #1598 .

fox0430 (approved):

PR #1597, opened 2023-01-23
[Enhancement] Add Shell Syntax Highlighting

### LINKED ISSUE #422: Supports shell script

kevinmatthes:
@fox0430

Fixes #422.  The syntax highlighting is based on the Python configuration with some minor enhancements.


kevinmatthes:
@fox0430

I tried to implement the highlighting for common Bash configuration files, as well, but they are not recognised due to their naming convention like `.bash_aliases` and so on.

The basic configuration ships with common shell language keywords, especially those for Bash, as well as some common non-Bash keywords in order to support a wide range of shell languages.  Some from the latter group of keywords are usually brought to Bash by the `core-utils` package such that they can be considered quasi-keywords.

src/moepkg/syntax/highlite.nim
```diff
@@ -229,19 +229,20 @@ proc isKeyword*(x: openArray[string], y: string): int =
...
+  of langNim: g.nimNextToken
   of langNone: assert false
-  of langPython: pythonNextToken(g)
-  of langRust: rustNextToken(g)
-  of langYaml: yamlNextToken(g)
+  of langPython: g.pythonNextToken
+  of langRust: g.rustNextToken
+  of langShell: g.shellNextToken
+  of langYaml: g.yamlNextToken
   else: discard
```
kevinmatthes, line 247: @fox0430

The compiler considers this `else` unreachable as all languages are redirected appropriately.  Before making changes to this section, I would like to discuss possible options in order to avoid the introduction of bugs.  Maybe the assertion in line 243 could be omitted, for instance?

fox0430: @kevinmatthes 

I think might as well omit that one. Can you omit it?

kevinmatthes: @fox0430

Okay, I will remove line 243. :+1:

fox0430 (approved):
Looks great. Thank you for your contribution.

PR #1608, opened 2023-02-02
[Enhancement] Add `:bg` and `:man` Commands

### LINKED ISSUE #1602: [Discussion] Special Use Cases of Shell Command Execution
@fox0430

I found the following two use cases of the `:!` command being pretty common in my development workflow:
- `:!echo` to see the recent terminal output, for example from a `grep` command for lines to edit
- `:!man <something>` to view a certain manual page, especially for C standard library functions

Both can be made easily into dedicated commands on their own as they are special cases of the regular command execution.  The first one is simply `shellCommand` without line 1085 (https://github.com/fox0430/moe/blob/develop/src/moepkg/exmode.nim#L1085) and could be named `bg`, for example, as it is the opposite action to the UNIX command `fg`.  Although Moe explicitly supports `Ctrl + Z` out of the box, a dedicated internal command might avoid forgetting about a paused Moe instance in the background.

The second one is another special case of `shellCommand` without the lines 1086 and 1087 (https://github.com/fox0430/moe/blob/develop/src/moepkg/exmode.nim#L1086; https://github.com/fox0430/moe/blob/develop/src/moepkg/exmode.nim#L1087) as the editor perspective would be restored automatically when `man` returns.  The command could be named `man` and take the arguments for `man` as parameters.  As it is already officially recommended to run Moe in a Linux environment, the presence of the `man` executable could be considered a sane assumption.

I would like to discuss whether it would be useful to give those two use cases commands on their own.

kevinmatthes:
@fox0430

This Pull Requests fixes #1602.

- The new commands are implemented.
- The new commands are documented in the help.
- The new commands are documented in the command suggestion.
- The new commands are documented in the "How to Use" command overview.

I just need to update my Scriv fragment, then, the PR is ready for merging.

### DIFF
--- a/src/moepkg/commandlineutils.nim
+++ b/src/moepkg/commandlineutils.nim
@@ -8,11 +8,12 @@ type
     filePath
 
 # TODO: Auto inserts spaces in compile time.
-const exCommandList: array[65, tuple[command, description: string]] = [
+const exCommandList: array[67, tuple[command, description: string]] = [
   (command: "!", description: "                    | Shell command execution"),
   (command: "deleteParen", description: "          | Enable/Disable auto delete paren"),
   (command: "b", description: "                    | Change the buffer with the given number"),
   (command: "bd", description: "                   | Delete the current buffer"),
+  (command: "bg", description: "                   | Pause the editor and show the recent terminal output"),
   (command: "bfirst", description: "               | Change the first buffer"),
   (command: "blast", description: "                | Change the last buffer"),
   (command: "bnext", description: "                | Change the next buffer"),
@@ -45,6 +46,7 @@ const exCommandList: array[65, tuple[command, description: string]] = [
   (command: "log", description: "                  | Open the log viewer"),
   (command: "ls", description: "                   | Show the all buffer"),
   (command: "lsw", description: "                  | Show the all workspace"),
+  (command: "man", description: "                  | Show the given UNIX manual page, if available"),
   (command: "multipleStatusLine", description: "   | Enable/Disable multiple status line"),
   (command: "new", description: "                  | Create the new buffer in split window horizontally"),
   (command: "noh", description: "                  | Turn off highlights"),
--- a/src/moepkg/exmode.nim
+++ b/src/moepkg/exmode.nim
@@ -1122,8 +1122,6 @@ proc manualCommand(status: var EditorStatus, manualInvocationCommand: string) =
   else:
     status.commandLine.clear
 
-  restoreTerminalModes()
-  status.commandLine.clear
   status.changeMode(currentBufStatus.prevMode)
 
 proc listAllBufferCommand(status: var EditorStatus) =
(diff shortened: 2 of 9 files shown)


src/moepkg/exmode.nim
```diff
@@ -237,6 +237,13 @@ proc isForceQuitCommand(command: seq[seq[Rune]]): bool {.inline.} =
 proc isShellCommand(command: seq[seq[Rune]]): bool {.inline.} =
   return command.len >= 1 and command[0][0] == ru'!'
 
+proc isBackgroundCommand(command: seq[seq[Rune]]): bool {.inline.} =
+  return command.len == 1 and cmpIgnoreCase($command[0], "bg") == 0
+
+proc isManualCommand(command: seq[seq[Rune]]): bool {.inline.} =
+  # TODO:  Configure a default manual page to show on `:man`.
+  return command.len > 1 and cmpIgnoreCase($command[0], "man") == 0
```
kevinmatthes, line 245: @fox0430

`command.len >= 1` allows for `:man` and thus the configuration of a default manual page to show.  I would like to suggest to either redirect to `man`'s manual page or Moe's one (as soon as it was written).

This is opted out for now but can be changed very easily.

src/moepkg/exmode.nim
```diff
@@ -1091,6 +1098,26 @@ proc shellCommand(status: var EditorStatus, shellCommand: string) =
...
+  restoreTerminalModes()
+  status.commandLine.clear
+  status.changeMode(currentBufStatus.prevMode)
+
+proc manualCommand(status: var EditorStatus, manualInvocationCommand: string) =
+  saveCurrentTerminalModes()
+  exitUi()
+
+  # TODO:  Configure a default manual page to show on `:man`.
+  discard execShellCmd(manualInvocationCommand)
```
kevinmatthes: @fox0430

Here, the default manual page can be configured.  If the `manualInvocationCommand` equals `man`, the default page could be shown.  Else, there are arguments to be passed to `man`.

src/moepkg/commandlineutils.nim
```diff
@@ -45,6 +46,7 @@ const exCommandList: array[65, tuple[command, description: string]] = [
   (command: "log", description: "                  | Open the log viewer"),
   (command: "ls", description: "                   | Show the all buffer"),
   (command: "lsw", description: "                  | Show the all workspace"),
+  (command: "man", description: "                  | Show the given UNIX manual page, if available"),
```
kevinmatthes, line 49: @fox0430

`man` will write an error message to the terminal if the queried page does not exist.  The user shall be responsible for calling only existing manual pages.

src/moepkg/exmode.nim
fox0430: If it fails, I think it's better to display an error message.

```suggestion
  let exitCode = execShellCmd(manualInvocationCommand)
  restoreTerminalModes()

  if exitCode != 0:
    let mess = "Error: No manual entry for " & manualInvocationCommand
    status.commandLine.writeMessageOnCommandLine(mess, EditorColorPair.errorMessage)
    status.messageLog.add(mess.toRunes)
  else:
    status.commandLine.clear
```

fox0430 (changes requested):

src/moepkg/exmode.nim
kevinmatthes: @fox0430

I applied the requested changes such that the error message will now be shown on the editor's command line, as well. :wrench: :heavy_check_mark:

fox0430 (approved):
Thank you for your contribution!

PR #1610, opened 2023-02-02
Fix #1609

### LINKED ISSUE #1609: [Bug Report] Discarded `:` in Help Buffer's Ex Mode
@fox0430

This is a minor visual bug.

With the latest available repository state of Moe, open any file.  Now, call `:help` and switch to the help buffer.  If the help buffer is active, the initial `:` in Ex Mode will be discarded visually, and only visually; any command will work as usual.  This only affects the help buffer if it is opened and active.

fox0430:
Fix a bug in that the prompt is not displayed when changing the mode from help to ex.

Close #1609

### DIFF
--- a/changelog.d/20230203_081121_shuu.n_fix_1609.rst
+++ b/changelog.d/20230203_081121_shuu.n_fix_1609.rst
@@ -1,3 +1,4 @@
+.. _#1609: https://github.com/fox0430/moe/pull/1609
 .. _#1610: https://github.com/fox0430/moe/pull/1610
 .. A new scriv changelog fragment.
 ..
@@ -21,7 +22,7 @@
 Fixed
 .....
 
-- `#1610`_ Fix #1609
+- `#1610`_ Fix `#1609`_
 
 .. Removed
 .. .......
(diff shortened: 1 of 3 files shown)


tests/thelp.nim
```diff
@@ -1,13 +1,32 @@
...
+import std/[unittest, strutils, importutils]
+import moepkg/[unicodeext, gapbuffer, editorstatus, bufferstatus]
 
 import moepkg/helputils {.all.}
+import moepkg/help {.all.}
+import moepkg/commandline {.all.}
 
-suite "Help":
+suite "initHlepModeBuffer":
   test "initHlepModeBuffer":
```
kevinmatthes: @fox0430

Please fix the typo `Hlep` (`e` and `l` exchanged) in lines 8 and 9.

```suggestion
suite "initHelpModeBuffer":
  test "initHelpModeBuffer":
```

changelog.d/20230203_081121_shuu.n_fix_1609.rst
```diff
@@ -0,0 +1,35 @@
...
+..
+.. Deprecated
+.. ..........
+..
+.. - A bullet item for the Deprecated category.
+
+Fixed
+.....
+
+- `#1610`_ Fix #1609
```
kevinmatthes: @fox0430

How about also linking the issue?

kevinmatthes (changes requested):
@fox0430

Thank you for your fix of the described bug!  I had a look at the changes and would like to suggest two small things explained below.

kevinmatthes (approved):

PR #1612, opened 2023-02-04
Fix #1611

### LINKED ISSUE #1611: [Bug Report] Calling Suggestions on Paths with Typos can Crash on Opening Files with `:e`
@fox0430

This is a rather serious bug, in my opinion.

The best description is the set of instructions how to reproduce:

1. Create a new directory somewhere, for example `bug/` and go there.
2. With `bug/` as your working directory, create a new directory `src/` as a sub-directory of `bug/`.
3. Create two new files, say `source1.nim` and `source2.nim` in `src/`.
4. With `bug/` as your working directory, call `moe src/source1.nim`.
5. With or without making changes to `src/source1.nim`, call `:e src/soru` and press the tab key once to try to get the suggestion for `src/source2.nim`.  This should crash the editor.  Please make source to intentionally type the name of the intended file wrong to simulate the typo `src/soruce2.nim` instead of the actual intention `src/source2.nim` (`o` and `r` are exchanged).

In a common development workflow, users often open already existing files to edit them.  As typos may happen in the source code, typos can be entered unintentionally in the path description to open a file, as well.  To reduce the possibility of spelling a path wrong, the suggestion feature of Ex Mode is quite useful.  It is common to enter the first few characters to reduce the length of the suggestion list.  If a rather simple typo breaks the session, the unsaved changes to the opened buffers will be lost.  Thus, I consider this bug serious.

fox0430:
Fix a bug that calling suggestions on paths with typos can crash on opening files with `:e`.

Close #1611

### DIFF
--- a/changelog.d/20230205_023217_shuu.n_fix_1611.rst
+++ b/changelog.d/20230205_023217_shuu.n_fix_1611.rst
@@ -1,0 +1,37 @@
+.. _#1611: https://github.com/fox0430/moe/pull/1611
+.. _#1612: https://github.com/fox0430/moe/pull/1612
+
+.. A new scriv changelog fragment.
+..
+.. Uncomment the header that is right (remove the leading dots).
+..
+.. Added
+.. .....
+..
+.. - A bullet item for the Added category.
+..
+.. Changed
+.. .......
+..
+.. - A bullet item for the Changed category.
+..
+.. Deprecated
+.. ..........
+..
+.. - A bullet item for the Deprecated category.
+
+Fixed
+.....
+
+- `#1612`_ Fix `#1611`_
+
+.. Removed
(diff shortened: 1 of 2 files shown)


kevinmatthes:
@fox0430

It seems to fix the bug but also to disable the suggestion window on the command line...

kevinmatthes:
@fox0430

Sorry, I am wrong.  Anything alright.

kevinmatthes:
@fox0430

I tried to reproduce the changes for testing and forgot one line.  Now, it works properly.

fox0430:
@kevinmatthes

Okay, thank you for the bug report and cooperation in fixing it!

kevinmatthes (approved):

PR #1629, opened 2023-02-26
[Bugfix] Node 12 Warning

kevinmatthes:
@fox0430

This PR fixes the warning regarding the usage of Node 12.

So far, the Scriv fragments were pushed to the repository using a certain Action which still uses the deprecated version 12 of Node.js.  Doing so triggered a warning in the Action overview.  That Action does not need to be used anymore if the settings of the other Actions in the respective workflows are adjusted appropriately.


fox0430 (approved):
Thank you!

PR #1630, opened 2023-02-26
[Bugfix] Fix Odd Edge Case of Markdown Comments

kevinmatthes:
@fox0430

There is a literally odd edge case of Markdown comments:  `--->` for termination.  By now, the lexer demands an even number of dashes for a Markdown comment to terminate.  But in practice, there might also be odd numbers of dashes to end a long Markdown comment.  This PR fixes this problem.

For illustration of the bug, please create a Markdown file, say, `test.md`, and add the following lines to it:

```markdown
<!--->...
<!---->...
<!----->...
```

The current version of Moe will consider the first and third lines unterminated while actually, only the first one is.  By adding one further dash (`-`) before the `>` in line three, the comment will be terminated.  Each subsequent dash will toggle the comment termination of the third line, indicated by the highlighting colour of the three dots at the end of each line.


fox0430 (approved):
Nice work. Thank you.

PR #1707, opened 2023-05-04
feat: Add a CLI option for init the config file

### LINKED ISSUE #1706: [Feature] Default Configuration File Generation Option
### Detailed Description

I would appreciate an option to the executable due to which the current user's configuration would be overwritten with the default configuration of Moe.  Here are some example use cases:

- users who have heavily changed the configuration and intend to reset their settings to the official default values
- users of nightly builds who would like to have access to the latest changes which require multiple changes to the configuration file

Some other editors have a similar feature, as well; Zee, for instance, calls this option `--init`.  The realisation would only require to dump the configuration into the configuration file, thereby truncating and, thus, overwriting all former contents of the file.

fox0430:
- Add `--init` CLI option
- Create/Overwrite the default configuration file and exit

Close #1706 

@kevinmatthes I implementd #1706. Do you have any requests?

### DIFF
--- a/changelog.d/20230505_052950_shuu.n_add_init_option_for_config.rst
+++ b/changelog.d/20230505_052950_shuu.n_add_init_option_for_config.rst
@@ -1,0 +1,37 @@
+.. _#1706: https://github.com/fox0430/moe/issues/1706
+.. _#1707: https://github.com/fox0430/moe/pull/1707
+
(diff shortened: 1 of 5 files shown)


kevinmatthes (approved):

kevinmatthes:
@fox0430

Looks great, thank you!

PR #1723, opened 2023-05-22
Enhancement: Add comment-changes workflow

### LINKED ISSUE #1722: [Discussion] Scriv Fragment Creation
@fox0430

In the Scriv repository, a user suggested the introduction of an interactive mode for `scriv create` such that one could select the category for which one is about to introduce changes.  This would be a great chance to minimise the file size of the resulting fragments but I am also afraid that this new mode might replace the current behaviour of being non-interactive such that any automation relying on the non-interactive behaviour would break.  This would also include the Scriv fragment creation Action this repository uses.  Although I realise that the pure suggestion of an interactive mode does not necessarily imply its implementation, I would like to discuss the possible options with you how the resulting issues might be best avoided, anyway.

One option might be to pin the Action to the latest version of Scriv which still uses the non-interactive behaviour but this might lead to vulnerabilities in the long term as I would rely on a version which is no longer maintained.  For this reason, that should not be preferred solution.

Instead, I created a workaround for the creation of a Scriv fragment.  The workaround is general enough to not necessarily be limited to Scriv.  The approach is that the desired fragment is generated from the commit messages of the current branch.  The application implementing this workaround reads the commit messages and extracts the items for the fragment which is created in a way compatible with Scriv.  Thus, the `scriv collect` step in the release preparation workflow would still work without any trouble.  Furthermore, one does not need to take care about the fragment's contents anymore as they are generated from the commit messages which is a great improvement regarding time efficiency.

In general, would you be interested in giving this workaround a try?

fox0430:
Close #1722


.github/workflows/comment-changes.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Checkout source
+        uses: actions/checkout@v3.5.2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.head_ref }}
+
+      - name: Setup git user
+        uses: fregante/setup-git-user@v2.0.1
+
+      - name: Install nightly toolchain
```
kevinmatthes: Hosted GitHub Action workflow runners are shipped with Rust installed; in fact, they even have GCC and the GNU Fortran compiler installed.  Thus, this step can be abbreviated to a simple `rustup update` because

- `rustup` is already installed on the runner.
- Aeruginous does not require Nightly Rust, Rust Stable suffices.

However, Aeruginous always requires the latest available version of Rust Stable due to requiring the access to the very latest features of Rust to offer all configured modes.  During the development, I noticed that the GitHub hosted runners have a delay of about one week before applying the very latest Rust toolchain version.  Hence, a call to `rustup update` is definitely required as the Aeruginous repository automatically upgrades its required version of Rust once per week.

changelog.d/20230522_220116_GitHub_Actions_add-aeruginous-workflow.rst
kevinmatthes, line 1: This fragment can be deleted due to not matching the Scriv settings; "Enhancement" is not configured to be a CHANGELOG category, yet.  Some hours ago, I implemented the functionality to create links such that you can run the following commands to receive a perfect Scriv fragment:

```bash
git commit --allow-empty -m 'Added: `#1723`_ comment-changes workflow; fixes `#1722`_'
aeruginous comment-changes -d : -n 1 -o changelog.d/ -l \#1722 -t https://github.com/fox0430/moe/issues/1722 -l \#1723 -t https://github.com/fox0430/moe/pull/1723
git push
```

The new options `-l` and `-t` represent `--link` and `--target`, respectively, such that you can define custom links in the fragment.  As you can see, you are allowed to define multiple links that way; the first given link name will be associated with the first given link target, the second name with the second target, and so on.  If there should be more link names than targets or vice versa, the arguments which cannot be associated anymore will be ignored silently.  (Or should I make Aeruginous print a warning, then?)

Please note that this will create an empty commit, i.e., it will not contain any changes.  You can also name some change to this branch with the above commit message and run Aeruginous on it to achieve the same result.

changelog.d/20230523_065316_shuu.n_add_aeruginous_workflow.rst
kevinmatthes, line 1: Please see my comment from above to create a smaller fragment with the very same effect.

.github/workflows/comment-changes.yml
```diff
@@ -0,0 +1,60 @@
...
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          rustup install nightly
+
+      - name: Install aeruginous-rs
+        run: |
+          cargo install --git https://github.com/kevinmatthes/aeruginous-rs
+
+      - name: Create commit changes
+        run: |
+          aeruginous comment-changes -d ':' -n 1 -o changelog.d/
```
kevinmatthes: I finally found out how to get the PR number, even *without* having to call any Action!

```bash
aeruginous comment-changes -d : -n 1 -o changelog.d/ -l ${{ github.event.pull_request.number }} -t https://github.com/fox0430/moe/pull/${{ github.event.pull_request.number }}
```

kevinmatthes (commented):
I added some comments on the new features I implemented into Aeruginous; you can now create links within the resulting fragment!  I also added the semantics to query the PR number.

Please check whether these suggestions would work for you.

kevinmatthes:
@fox0430

Aeruginous v1.1.0 is released! :tada:

This means that you can fetch the application from the Rust registry as a stable binary by calling `cargo install aeruginous`.  If there should be changes to the calling semantics in future versions, I will let you know such that the workflow will not break.

kevinmatthes (changes requested):
@fox0430

The workflow is perfect, now!

As a last step, I would like to ask to you to create an *empty* commit on this branch using the following command.

```bash
git commit --allow-empty -m 'Added: `#1723`_ comment-changes workflow; fixes `#1722`_'
```

Then, please go the Actions tab on the GitHub website and manually trigger the `comment-changes` workflow on this branch.  This will be the final test for the workflow before it will be applied in production.

fox0430:
@kevinmatthes 

I can not trigger the workflow... Do you know how to trigger it before merging? Do I need to reopen this PR..?

kevinmatthes:
@fox0430

The easiest would be to:

1. add `reopened` as trigger condition between line 25 and 26 in the workflow file and commit this change with the message ```Added: `#1723`_ comment-changes workflow```.
2. close this PR.
3. reopen this PR.
4. wait for the workflow to succeed before removing the trigger condition `reopened` in the workflow file, again.

This should solve the problem.

changelog.d/20230524_171657_GitHub_Actions_add-aeruginous-workflow.rst
kevinmatthes, line 1: You could either add the link to the issue or remove the part ```; fixes `#1722`_```.  I would recommend the latter option as the workflow by default will only create links to the PR.  GitHub will take care about metadata annotation by mentioning the PR number, so it will be implicitly clear to visitors that this PR is related to #1722 as this is outlined in the PR's opening comment.

kevinmatthes (changes requested):
Wonderful!

There is one more thing left regarding the rendered fragment; see below.  Please change this last detail and I will give my approval to this PR.

I would like to suggest to leave the alternative trigger condition of the workflow in; it might be useful in the future.

changelog.d/20230524_171657_GitHub_Actions_add-aeruginous-workflow.rst
fox0430: @kevinmatthes 

Okay, I remove it. But is it impossible to create an issue link? I want to use commit messages like this `Fix #xxx; description`.

kevinmatthes: @fox0430

It is not completely impossible but difficult.  You would need to extract the issue number from the commit message and pass it to Aeruginous.  This is dangerous as it would allow for code injection.

Aeruginous itself allows for the linking to issues but we would need to implement this into the workflow somehow.  I will keep looking for a solution to this.  For now, I would like to suggest to merge the current design of the workflow as a basic setup for further adjustments later on.

fox0430: @kevinmatthes 

Okay, Thank you for your suggestion and support!

kevinmatthes (approved):

PR #1727, opened 2023-05-24
[Bugfix] Fix #1555

### LINKED ISSUE #1555: [Bug Report] Cursor does not Change Back to Normal
@fox0430

When editing some files with Moe in the KDE terminal `konsole`, v20.12.3, I start with a solid cursor (i.e., it is not blinking).  If Moe is allowed to change cursors, the cursor will start to blink as soon as the editor runs.  When quitting Moe regularly by using one of `:WQ`, `:Q`, and `:Q!`, the cursor still keeps blinking after the editor is left.  Changing back to the normal solid cursor requires restarting the terminal.

I consider this a bug since the expectation would be that the cursor will only blink within the editor and returns to solid state after leaving it.

(I apologise for this late bug report; I disabled the cursor changing when I started to use Moe and forgot to report this bug.  I am sorry.)

fox0430:
Restore the cursor shape when quitting moe using `x1B[0 q`.

This solution is incomplete and I don't know how many terminals support it. It's maybe not possible at the moment to restore the terminal cursor shape in all environments.

### DIFF
--- a/changelog.d/20230524_210000_GitHub_Actions_fix#1555.rst
+++ b/changelog.d/20230524_210000_GitHub_Actions_fix#1555.rst
@@ -1,0 +1,2 @@
+.. 1727:  https://github.com/fox0430/moe/pull/1727
+
--- a/src/moepkg/ui.nim
+++ b/src/moepkg/ui.nim
@@ -117,7 +117,7 @@ proc changeCursorType*(cursorType: CursorType) =
     of terminalDefault: setTerminalDefaultCursor()
     of blinkBlock: setBlinkingBlockCursor()
     of noneBlinkBlock: setNoneBlinkingBlockCursor()
-    of blinkIbeam: setBkinkingIbeamCursor()
+    of blinkIbeam: setBlinkingIbeamCursor()
     of noneBlinkIbeam: setNoneBlinkingIbeamCursor()
 
 proc disableControlC*() {.inline.} =
--- a/src/moepkg/ui.nim
+++ b/src/moepkg/ui.nim
@@ -82,7 +82,7 @@ proc getTerminalHeight*(): int = terminalSize.h
 
 proc getTerminalWidth*(): int = terminalSize.w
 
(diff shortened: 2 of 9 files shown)


changelog.d/20230524_210000_GitHub_Actions_fix#1555.rst
kevinmatthes, line 1: This fragment is empty because the workflow was triggered after the content delivering commit was pushed.

Furthermore, it is necessary to adjust https://github.com/fox0430/moe/blob/develop/.github/workflows/comment-changes.yml#L61 to `-l \#${{ github.event.pull_request.number }}`; the key change is the `\#` because the link name contains a hash character.  The missing underscore in line 1 in the resulting fragment is my fault because I forgot it in Aeruginous; I will fix it and publish a patch release, soon.

I will also add a category checker and configure it in the workflow; `Bugfix` is not configured to be a Scriv category, yet, so I will implement a catching step into Aeruginous in addition to the above bugfix.

You can leave the fragment as it is, for now, I will take care about the changes to all affected fragments when I published the patch for Aeruginous.

kevinmatthes (approved):
@fox0430

You can go ahead and merge these changes.  As I wrote, I am going to take care about the fragments.

src/moepkg/ui.nim
```diff
@@ -81,22 +82,43 @@ proc getTerminalHeight*(): int = terminalSize.h
 
 proc getTerminalWidth*(): int = terminalSize.w
 
-proc setBkinkingIbeamCursor*() {.inline.} = discard execShellCmd("printf '\e[5 q'")
+proc setBkinkingIbeamCursor*() {.inline.} =
```
kevinmatthes:
```suggestion
proc setBlinkingIbeamCursor*() {.inline.} =
```

fox0430:
Link #1555

kevinmatthes (approved):

PR #1728, opened 2023-05-25
[Bugfix] Adjust `comment-changes` Workflow

kevinmatthes:


kevinmatthes:
@fox0430

There is still a bug but this time not in Aeruginous; the workflow works only with *this* repository because it only checks out this one.  But at least, it now works as it should, i.e. with the underscore before the link names and the Scriv category filters applied.

kevinmatthes:
@fox0430

It now checks out the correct repository but it is not allowed to push commits, even with "Allow edits by maintainers" set.  But the settings are now correct for using it in this repository.

fox0430 (approved):

PR #1732, opened 2023-05-28
Enhancement: Attempt to speed up cargo installs using caching in CI

fox0430:


fox0430:
@kevinmatthes 

Fragments only creates links, is my commit message wrong?

kevinmatthes:
@fox0430

Unfortunately, yes, your commit message is wrong.  The CHANGELOG categories are configured to be the following:

- Added
- Changed
- Deprecated
- Fixed
- Removed
- Security

Your commit message starts with "Enhancement" which is not configured to be a category.

The filter option to the fragment creation shall make sure that such issues can be identified and fixed early, i.e., directly when they are arising; an empty fragment is more obvious than a not recognised one which was skipped on assembly because Scriv did not found any category information.

fox0430:
@kevinmatthes 

I understand. Thank you. I will use these categories from the next PR.

PR #1737, opened 2023-06-06
[Bugfix] Always Install Latest Aeruginous Version

kevinmatthes:
The current setup of the `cargo-install` Action always grabbed statically v1.1.2 instead of fetching *at least* that version of Aeruginous.  Thus, some bugfixes would not be available.  By removing the version field, the respective latest version will be downloaded ensuring access to all improvements and fixes of this mode.  Furthermore, some obsolete dependencies of Aeruginous were removed with v2.0.0 such that the compilation in general should be faster, now.

### DIFF
--- a/.github/workflows/comment-changes.yaml
+++ b/.github/workflows/comment-changes.yaml
@@ -53,7 +53,6 @@ jobs:
         uses: baptiste0928/cargo-install@v2
         with:
           crate: aeruginous
-          version: "^1.1.2"
 
       - name: Create commit changes
         run: |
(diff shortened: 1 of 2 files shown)


fox0430 (commented):
@kevinmatthes 

Thank you for PR.

The workflow failing,  but can I safely ignore it in this PR?

kevinmatthes:
@fox0430

It can be safely ignored; see the run details and exit cause for details.  The reason is that the GitHub Actions bot will attempt to push a commit into my fork during the workflow run.  This is impossible due to security reasons.  The workflow will work well for you because you can think of it as *your* instance of the GitHub Actions bot, thus, there will not be any problems for you.

fox0430 (approved):

PR #1750, opened 2023-07-07
Fixed:  crash due to typo in example configuration

kevinmatthes:
There is a small typo in the example configuration which causes the editor to crash.  This PR fixes that typo.


fox0430 (approved):
Thank you!

PR #1752, opened 2023-07-07
feat: Add `colorMode` setting

fox0430:
Add a setting for what color mode to use.

- Add `colorMode` to Standard table
- `none` or `8`  or `16` or `256` or `24bit`


kevinmatthes:
@fox0430

Thank you, this patch solves the problem for me!

PR #1771, opened 2023-07-18
[Bugfix] Prevent comment-changes Workflow from Failing

kevinmatthes:
This PR shall prevent the `comment-changes` workflow from failing by running it only if the repository the PR originates from is *this* repository.  Furthermore, I removed the manual trigger as the PR number is always required but a manual dispatch does not necessarily ensure that information.


fox0430 (approved):
Thank you!

PR #1775, opened 2023-07-21
[Enhancement] Switch to Maintenance-free cffconvert Docker Image

kevinmatthes:
The CFF validation Action is built upon a Docker image which is a wrapper around the Python 3 CLI.  We can reduce the maintenance effort of the CFF validation workflow by calling that Docker image directly with `latest` as requested version to always have access to the latest version of the validator.


fox0430 (approved):
Nice work.

PR #1776, opened 2023-07-21
[Enhancement] Add Linter for GHA Workflows

kevinmatthes:
This PR adds a linter for GitHub Action workflows which is called whenever workflow files were edited.  This shall ensure the correctness of the workflows and prevent difficult bug tracing.

If changes to the workflows should be necessary as the linter could complain about certain style issues, I will take care about those.


fox0430 (approved):
Great! Thank you!

PR #1868, opened 2023-09-28
feat: Auto split view when starting

### LINKED ISSUE #1867: [Feature Request] Automatic Split View when Invoked with Multiple Paths
### Detailed Description

At the moment, a command like `moe a.nim b.nim c.nim` will cause the editor to only open the last given path (`c.nim`).  I would like to ask for a new feature such that, if multiple paths should be specified, Moe would open them optionally in multiple sub-windows as if the user would have typed something like:

1. `moe a.nim`
2. `:vs`
3. Move to right sub-window.
4. `:e b.nim`
5. `:vs`
6. Move to right sub-window.
7. `:e c.nim`

This behaviour could be named `startUpAutoSplit` and should be configurable such that the user might decide whether the current behaviour shall be kept (e.g. `legacy`), or the window shall be subdivided `horizontally` (`:sv` command) / `vertically` (`:vs` command).  I would like to suggest defining `vertically` (`:vs` command) being the default as I assume that most screens are rather in landscape than in portrait mode, and other editors like Kate have a similar behaviour when passing multiple paths to the application, i.e. Kate would open each one in a separate tab.

fox0430:
Add settings to display all buffers in multiple views if multiple paths are received when starting the editor.

- Add `StartUp.FileOpen.autoSplit` setting
- Add `StartUp.FileOpen.splitType` setting

### TODO
- [x] Config mode
- [x] Tests
- [x] Docs

Close #1867

### DIFF
--- a/changelog.d/20230928_200227_GitHub_Actions_auto-split-open.rst
+++ b/changelog.d/20230928_200227_GitHub_Actions_auto-split-open.rst
@@ -3,5 +3,5 @@
 Added
 .....
 
-- `#1867`_ feat: Auto split view when starting
+- `#1868`_ feat: Auto split view when starting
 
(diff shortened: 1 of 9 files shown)


fox0430:
@kevinmatthes 

Please let me know if you have any opinions.

kevinmatthes:
@fox0430

It works perfectly!  Thank you!

PR #1877, opened 2023-10-06
feat: Add Language Server Protocol support

### LINKED ISSUE #610: Initial support for LSP

fox0430:
Add LSP client support. It's an experimental and incomplete feature.
I'm testing with [nimlsp](https://github.com/PMunch/nimlsp) but it probably still doesn't work on many servers.

- Add LSP client
- Add settings for LSP
- Add `K` command (Hover) to Normal mode

### Screenshots

lsp

### Supported LSP Commands
- `Initialize`
- `shutdown`
- `workspace/didChangeConfiguration`
- `textDocument/didOpen`
- `textDocument/didChange`
- `textDocument/didClose`
- `textDocument/hover`

Close #610 

## TODO
- [x] Send `textDocument/didChange` after changing buffers
- [x] Send `textDocument/didClose` after removing buffer
- [x] Check server capabilities
- [x] Error handling
- [x] Improve debug logs
- [x] Force enable debug logs when enabled LSP
- [x] Add immutable id
- [x] Fix that freeze when suggestions are not found
- [x] Fix popup window bugs
- [x] ~~Fix that too slow when sending `textDocument/didChange` at second time~~
- [x] Fix that cannot receive response from the server after timeout
- [x] Tests
- [x] help
- [x] Docs

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -84,7 +84,11 @@ jobs:
         cd abduco-0.6
         make && sudo make install
 
-    - name: Run unit test
+    - name: Install nimlsp
       run: |
         nimble refresh
+        nimble install nimlsp
+
+    - name: Run unit test
+      run: |
         xvfb-run nimble test --verbose -y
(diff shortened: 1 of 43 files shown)


fox0430:
The problem with slow sending when sending `didChange` twice quickly may be caused by nimlsp.
However, this occurs only once, so I will ignore it in this PR...

Issue #2028, opened 2024-03-01
Throws an exception on enter.


userwiths:
### Version

```
moe v0.3.0

Git hash: 89542793bbfed27e338a9fbdb1b0a415db90875f
Build type: Release
```

### Environment

Terminal : `alacritty`, using a [startship](https://starship.rs/) ran tests with and without `starship`, same result.
Tried with `konsole`, same result.

Running `X` display.

`uname -a` -> `Linux archlinux 6.7.6-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 23 Feb 2024 16:31:48 +0000 x86_64 GNU/Linux`

As per the moe requirements:
`extra/xclip 0.13-4`
`extra/xsel 1.2.1-1`
`core/ncurses 6.4_20230520-1`
`extra/nim 2.0.2-1`

### Category

Major — The software does not fulfill the task it is designed for.

### Steps to Reproduce

1. Startup `moe`
2. Click the `a` key once to enter insert/add mode.
3. Click `a` button a second time in order to write out an `a` sign on the screen.
4. Click enter.

### Detailed Description

The same error message (maybe for different reasons) can be seen in [this comment](https://github.com/fox0430/moe/issues/1881#issuecomment-1767056466).

`options.nim(230) get Error: unhandled exception: Can't obtain a value from a 'none' [UnpackDefect]`

![image](https://github.com/fox0430/moe/assets/32035685/e4431ae6-0b18-456b-b597-14aa55d4ac43)

userwiths:
Tested after running `nimble install moe` with the following version information

```
moe v0.3.0

Git hash: 22920ff1e8ff97ffa3acf2da326adbec10825d95
Build type: Release
```

Issue is fixed.

PR #2029, opened 2024-03-01
fix: Add safe guards to allow the editor to proceed.

userwiths:
Due to the fact that the console breaks on the exceptions, I find it hard to post exact exception messages. See pictures for detailed exception.

## Changes in `mainloop.nim`

As outlined in #2028 

image

## The changes in `insertmode.nim`

I added them after fixing the one in `mainloop` and meeting another exception, in summary `Index out of bound` but I attach a picture with the whole exception.

image

This is my best guess at fixing those issues. Locally it behaves good now and allows me to write in it without a problem.

### DIFF
--- a/src/moepkg/insertmode.nim
+++ b/src/moepkg/insertmode.nim
@@ -137,6 +137,8 @@ proc insertToBuffer(status: var EditorStatus, r: Rune) {.inline.} =
          error err.error
 
 proc execInsertModeCommand*(status: var EditorStatus, command: Runes) =
+  if command.len == 0:
+      return
   let key = command[0]
 
   if isCtrlC(key) or isEscKey(key):
--- a/src/moepkg/insertmode.nim
+++ b/src/moepkg/insertmode.nim
@@ -138,7 +138,7 @@ proc insertToBuffer(status: var EditorStatus, r: Rune) {.inline.} =
 
 proc execInsertModeCommand*(status: var EditorStatus, command: Runes) =
   if command.len == 0:
-      return
+    return
   let key = command[0]
 
   if isCtrlC(key) or isEscKey(key):
--- a/src/moepkg/mainloop.nim
+++ b/src/moepkg/mainloop.nim
@@ -378,7 +378,6 @@ proc incrementalReplace(
 template isOpenCompletionWindowInCommandLine(
   status: EditorStatus,
   key: Rune): bool =
-
     status.completionwindow.isNone and
     currentBufStatus.isExmode and
     status.settings.standard.popupWindowInExmode and
@@ -853,7 +852,8 @@ proc editorMainLoop*(status: var EditorStatus) =
               status.confirmCompletion
               isClosedCompletionWindow = true
 
-        command.add key.get
+        if key.isSome:
+            command.add key.get
 
         let inputState = invokeCommand(
           currentBufStatus.mode,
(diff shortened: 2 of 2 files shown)


userwiths:
One thing. After the changes I made and used the editor some time. I notice that I need to hit `Enter` 2 times to make a new like.
Tried to solve this, but I dont know the project well enough yet. Would be happy to discuss.

src/moepkg/insertmode.nim
```diff
@@ -137,6 +137,8 @@ proc insertToBuffer(status: var EditorStatus, r: Rune) {.inline.} =
          error err.error
 
 proc execInsertModeCommand*(status: var EditorStatus, command: Runes) =
+  if command.len == 0:
+      return
```
fox0430: Please use 2 spaces indents

src/moepkg/mainloop.nim
```diff
@@ -853,7 +852,8 @@ proc editorMainLoop*(status: var EditorStatus) =
               status.confirmCompletion
               isClosedCompletionWindow = true
 
-        command.add key.get
+        if key.isSome:
+            command.add key.get
```
fox0430: Please use 2 spaces indents

src/moepkg/insertmode.nim
userwiths: Should be ok now.

src/moepkg/mainloop.nim
userwiths: Should be ok now.

fox0430:
@userwiths 

> One thing. After the changes I made and used the editor some time. I notice that I need to hit Enter 2 times to make a new like.
Tried to solve this, but I dont know the project well enough yet. Would be happy to discuss.

First, Thank you for your PR. This problem seems to be due to the completion window.
Probably, `status.completionWindow` is Some but suggestions are not found and no visual window is visible. However, since `status.completionWindow` is Some, Completion Window handling is enabled and we need to press Enter key twice.

That is, the first Enter key is used in the completion window but nothing happens in the above situation, and the second time is used to insert a character.

It might be a bit troublesome so I'll fix it in another PR.

fox0430:
@userwiths 

Please change line 833 ~ 853 in `mainloop.nim`. Add `status.completionWindow.get.popupWindow.isSome` to the condition.
Should be resolve the 2 enter key problem.
```nim
        if status.completionWindow.isSome and
           status.completionWindow.get.popupWindow.isSome:
             if canHandleInCompletionWindow(key.get):
               status.completionWindow.get.handleKey(
                 currentBufStatus,
                 currentMainWindowNode,
                 key.get)
               key.resetKeyAndContinue
             else:
               if isEnterKey(key.get):
                 status.confirmCompletion
                 key = none(Rune)
                 isClosedCompletionWindow = true
               elif isConfirmCompletionAndContinue(status, key.get):
                 status.confirmCompletionAndContinue
               elif isEscKey(key.get):
                 status.confirmCompletion
               elif isBackspaceKey(key.get):
                 status.confirmCompletionAndContinue
               elif not isCompletionCharacter(key.get):
                 status.confirmCompletion
                 isClosedCompletionWindow = true
```

userwiths:
Hey, I tried what you suggested but the suggestion window disappeared.

I did a little more searching and found out that really there are 0 suggestions in the `completionWindow` and added a check on the enter key to see if there are any suggestions.

fox0430:
@userwiths 

Great! Thank you!

fox0430 (approved):

PR #2038, opened 2024-03-30
(fix) directly call getEnv to get XDG_SESSION_TYPE

xTrayambak:
Moe currently invokes the shell to get whether the session type is Wayland or not. This is not necessary as `XDG_SESSION_TYPE` can be directly fetched using `getEnv` and we don't even need to call `contains`, we can directly equate it to "wayland" (lowercase) or "x11" (lowercase). This PR makes Moe now depend on `XDG_SESSION_TYPE` on X11 as well instead of evoking another program.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -80,4 +80,4 @@ jobs:
 
     - name: Run unit test
       run: |
-        xvfb-run nimble test --verbose -y
+        xvfb-run env XDG_SESSION_TYPE=x11 nimble test --verbose -y
--- a/src/moepkg/clipboard.nim
+++ b/src/moepkg/clipboard.nim
@@ -64,11 +64,10 @@ template wslDefaultPasteCommand(): string =
 template macOsDefaultPasteCommand(): string = "pbpaste"
 
 template isXAvailable*(): bool =
-  execCmdEx("xset q").exitCode == 0
+  getEnv("XDG_SESSION_TYPE") == "x11"
 
 template isWaylandAvailable*(): bool =
-  let r = execCmdEx("echo $XDG_SESSION_TYPE")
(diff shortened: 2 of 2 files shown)


fox0430 (approved):

fox0430:
@xTrayambak 

Looks like a good way. Thank you!

fox0430:
@xTrayambak 

It looks like unit tests are failing. Can you fix it?

xTrayambak:
Okay.

xTrayambak:
image
The test that's supposed to fail is working fine on my system for some reason and the changes and the changes I've made shouldn't even do anything to make them fail unless one of the functions somehow invokes the clipboard, which shouldn't break anything either ways.

fox0430:
@xTrayambak 

 Tests fail on xvfb. but I'm still not sure...
https://github.com/fox0430/moe/actions/runs/8498509413/job/23278292346

 In my environment, even if I run the following command, everything passes, so it seems that Actions is the cause.
```
xvfb-run nim c -r tests/teditors
```

xTrayambak:
I don't have X11 on my system at all, so that might be why the tests all passed for me, but I have no clue why Actions failed if it's fine with you.

fox0430:
@xTrayambak 

I found out the cause of this. The cause is that nothing is set to `XDG_SESSION_TYPE` on xvfb.
Please fix a workflow like this 
https://github.com/fox0430/moe/blob/d3bd047033ea68e42f1c56c0371afd58365cbe42/.github/workflows/test.yaml#L76

xTrayambak:
Sorry for the late response, I just fixed that.

fox0430 (approved):

fox0430:
@xTrayambak 

Thanks

PR #2140, opened 2024-08-18
ci: Disable Nim devel

fox0430:
Temporarily disable Nim devel on CI due to compiler bug


fox0430:
Ref https://github.com/nim-lang/Nim/issues/23998

PR #2170, opened 2024-09-02
ci: test with nightlies devel compiler

jiro4989:
## Overview

* use `use-nightlies` parameter in CI test with `devel` compiler
  * `use-nightlies` is available in [setup-nim-action `v2.2.0`](https://github.com/jiro4989/setup-nim-action/releases/tag/v2.2.0)
  * CI will complete faster because it will use [nightlies build](https://github.com/nim-lang/nightlies/releases) (latest-devel) compiler.


fox0430 (approved):

fox0430:
@jiro4989 

Great! Thank you!

PR #2745, opened 2026-06-09
fix(pty): fix freebsd include and add libutil link

lbartoletti:
On FreeBSD, pty is declared in `libutil.h`, not `util.h` and you have to link against it. Adding this `elif` fix to the FreeBSD build, and I can give a try to `moe`. Thanks


fox0430:
@lbartoletti

Thank you for your contribution. Can this be formatted with nph?

fox0430 (approved):

Issue #3020, opened 2026-07-18
Embedding in a GUI


elcritch:
### Detailed Description

Hey! I’m experimenting with putting Moe in my [Merenda GUI](https://github.com/elcritch/merenda) to make a new GUI editor. It builds on [FigDraw](https://github.com/elcritch/figdraw) which recently got support for code ligatures and such so it'd be fun to make a pure Nim GUI based editor.

Would you be open to small refactors that keep the current terminal UI unchanged but make Moe’s core frontend-neutral?

From reviewing the code it looks like AsyncApp in Editor would make embedding hard. Then there'd need to be KeyCombo/paste/mouse input entry points. Finally some way to render into a small Moe-owned cell grid with a Celina adapter? I have mono-font views that could work well with the cell grid style of Moe.

Perhaps those are things that could be added in Celina?

fox0430:
@elcritch 

Thank you for your interest in this project!
I happy to accept refactoring contributions. If adjustments are needed on the celina side as well, please feel free to open issues or PRs on celina.

elcritch:
Okay this PR changes from raw event handling to more semantic "key combo" handling and let's the frontend add a layer for raw events -> key combos. https://github.com/fox0430/moe/pull/3071

elcritch:
It's starting to work pretty well! 

Kosmo + Moe ❤️

<img width="1840" height="1191" alt="Image" src="https://github.com/user-attachments/assets/2254f7d2-cab0-4159-9fa5-681bd197dd0e" />

fox0430:
@elcritch 

Great news! Is it possible to add a CI that runs Kosmo's build on moe's CI so as not to break Kosmo's build?

elcritch:
> [@elcritch](https://github.com/elcritch)
> 
> Great news! Is it possible to add a CI that runs Kosmo's build on moe's CI so as not to break Kosmo's build?

That should be feasible! It's still changing too much right now but should be stabilizing soon.

elcritch:
BTW, thought this looked nice! The about page with the new kosmo icon:

<img width="396" height="356" alt="Image" src="https://github.com/user-attachments/assets/b65ae897-0bba-430f-a721-ef78dfd30199" />

PR #3074, opened 2026-07-22
refactor: remove celina app from editor state

elcritch:
## Summary
- Removes `AsyncApp` from `Editor` state.
- Queues frontend requests for mouse capture instead of calling Celina from editor config reload.
- Passes suspend/resume hooks into pending async operations so the app frontend owns terminal suspension.

This keeps the PR scoped to app/frontend ownership; generic drawing/render target work is left for a separate PR.

## Tests
- `nimble build --hints:off`
- `nimble c -r tests/test_editor_config_pull_accessors.nim`
- `nimble c -r tests/test_handler.nim`
- `nimble c -r tests/test_handler_manager.nim`


src/moepkg/command_handlers/editor_ops.nim
```diff
@@ -28,6 +28,8 @@ import std/[options, os, times, strutils, tables]
 
 import pkg/results
 
+import pkg/results
```
fox0430: Duplicate `import pkg/results`

src/moepkg/handler.nim
```diff
@@ -1236,12 +1236,28 @@ proc hasPendingAsyncOperations*(e: Editor): bool =
...
+    suspend*: FrontendSuspendHook
+    resume*: FrontendSuspendHook
+
   BuildInfo =
     tuple[path: string, language: int, customCmd: string, workspaceRoot: string]
   QuickRunInfo =
     tuple[cmd: string, args: seq[string], filePath: string, isTempFile: bool]
   SyntaxCheckInfo = tuple[path: string, language: int]
 
+template withFrontendSuspend(frontend: FrontendHooks, body: untyped) =
```
fox0430: When `suspend` is nil, the body still runs unprotected. The old `withSuspendAsync`
guaranteed the body never ran without a successful suspend (nil app -> exception),
but here the terminal stays in raw mode while the body does terminal I/O:
`readLine` waits for `\n`, but with ICRNL off Enter sends `\r`, and with ISIG off
Ctrl-C doesn't raise SIGINT - so neither works and the editor appears frozen.
(`:bg` similarly sends SIGTSTP without restoring the terminal first.)

Since the bodies are inherently terminal code (`execShellCmd`, `stdin.readLine`),
implementing suspend/resume in a GUI frontend wouldn't make this path safe either -
for the GUI embedding case (#3020) I think the right degradation is to skip the
body and report it:

```nim
template withFrontendSuspend(frontend: FrontendHooks, e: Editor, body: untyped) =
  if frontend.suspend.isNil or frontend.resume.isNil:
    e.state.statusMessage = "This frontend does not support terminal commands"
  else:
    await frontend.suspend()
    try:
      body
    finally:
      await frontend.resume()
```

Could you also drop the `= FrontendHooks()` default on `handlePendingAsyncOperations`
so every caller makes an explicit choice? Supporting `:!cmd` in a GUI would need a
different abstraction anyway (e.g. a shell-execution hook), which can be a follow-up.

elcritch: > Since the bodies are inherently terminal code (execShellCmd, stdin.readLine),
implementing suspend/resume in a GUI frontend wouldn't make this path safe either -
for the GUI embedding case (https://github.com/fox0430/moe/issues/3020) I think the right degradation is to skip the
body and report it:

Thanks! I wasn't quite sure about that bit. 

> Could you also drop the = FrontendHooks() default on handlePendingAsyncOperations

I can do that as well.

elcritch:
@fox0430 implemented the comments.

tests/test_editor_config_pull_accessors.nim
```diff
@@ -277,6 +275,19 @@ suite "Editor - applyConfigSettings live-reload (S1 regression)":
     check e.config.standard.lineWrap == true
     check e.lineWrap == true
 
+  test "mouse setting queues a frontend request":
+    let e = mkEditor()
+    discard e.state.takeMouseCaptureRequest()
```
fox0430: Add an assertion, please.
This is the request that enables mouse capture at startup, so it should be verified rather than discarded
```nim
check e.state.takeMouseCaptureRequest() == some(false)
```

fox0430 (changes requested):

fox0430 (approved):
Thanks!

PR #3164, opened 2026-07-29
refactor(input): add frontend-neutral GUI boundaries

elcritch:
Adds a frontend-neutral pointer/scroll API to Moe while keeping Celina as the terminal compatibility adapter. Scroll handling now returns the actual movement and precise content-grid region, excluding fixed UI such as tab and status lines. This enables GUIs to add scrolling animation.

With this PR GUIs can re-use Celina's Buffer types while still letting the GUI handle mouse input in addition to keyboards. This approach avoids a larger refactor to introduce a rendering bridge which adds a lot of complexity. This could be done in the future, but Celina Buffer seems to provide a good bridge for this currently.

### DIFF
--- a/src/moepkg/frontend_input.nim
+++ b/src/moepkg/frontend_input.nim
@@ -20,8 +20,9 @@
 ## Frontend-neutral pointer and scrolling input.
 ##
 ## Coordinates are cells in Moe's rendered grid. A GUI frontend should convert
-## view coordinates before constructing these values. Positive `deltaRows`
-## scrolls toward later buffer rows; negative values scroll toward earlier rows.
+## view coordinates before constructing these values. Positive
+## `deltaPhysicalRows` scrolls toward later buffer lines; negative values scroll
+## toward earlier lines.
 
 import key_bindings/registry
 export registry.KeyModifier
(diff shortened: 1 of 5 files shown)


src/moepkg/frontend_input.nim
```diff
@@ -0,0 +1,99 @@
...
+    clickCount*: Natural
+    modifiers*: set[KeyModifier]
+
+  ScrollInput* = object
+    row*: int
+    column*: int
+    deltaRows*: int
+    modifiers*: set[KeyModifier]
+
+  ScrollOutcome* = object
```
fox0430, line 65: The docstring promises a grid-row translation amount, but the value is `topLine - previousTopLine` - a physical-line delta sampled before layout. It diverges in 3 ways:

1. Line wrap: one physical line spans K grid rows, so the GUI's translation and the actual content shift disagree
2. Filer (`handler.nim:525`): `topLine` only moves later in `advanceLayoutForFrame`, so the value is always 0
3. Text-edit near the bottom (`handler.nim:560`): the inline keep-visible check uses raw `viewport.height`, but the render pass uses `height - steadyReservedLines - tabLineOffset` (`editor_render_views.nim:189`) — the PR's own test pins 3 where the layout settles `topLine` at 5. (Predates the PR; the new field just makes it observable.)

In every case the GUI translates by N while the frame shifts by a different amount, leaving stale rows until repaint.

```nim
  ScrollOutcome* = object
    ## Result of applying a semantic row scroll.
    ##
    ## Movement fields are signed physical-line deltas (positive = later rows);
    ## grid-row conversion under line wrap is the consumer's job.
    ## `viewportPhysicalRowsMoved` is a pre-layout approximation (may be 0 or
    ## short) — a repaint hint, not an exact translation.
    handled*: bool
    region*: GridRegion
    requestedRows*: int
    appliedRows*: int
    viewportPhysicalRowsMoved*: int
```

Consider renaming `ScrollInput.deltaRows` to `deltaPhysicalRows` for symmetry. Behavioral upgrades can wait for a real GUI consumer. The rename touches 5 sites in `tests/test_handler.nim`; wrap/Filer cases have no coverage.

src/moepkg/handler.nim
```diff
@@ -489,88 +473,104 @@ proc finalizeCurrentWindowForMouseJump(e: Editor) =
...
+    return
 
-        # Update viewport topLine to keep cursor visible
-        let viewportHeight = window.viewport.height
-        if viewportHeight > 0:
-          if newLine < window.viewport.topLine:
-            window.viewport.resetViewportTop(newLine)
-          elif newLine >= window.viewport.topLine + viewportHeight:
-            window.viewport.resetViewportTop(newLine - viewportHeight + 1)
+  result.handled = true
```
fox0430: The fallthrough marks every mode outside Filer and the text-editing set as handled although those modes implement no scrolling at all, while the early returns (zero delta, mouse disabled, no windows) report `handled = false` - two contradictory encodings of "scroll not applicable", so a GUI bridge cannot tell an inert mode from ignored input when deciding whether to fall back to native scrolling. (Boundary-clamped scrolls are a separate, legitimate case: `handled = true` with `appliedRows = 0`) The sibling handlePointerInputCore already returns false for unsupported modes, so the fallthrough is also asymmetric with the pointer path.

Fix: drop the fallthrough so inert modes (LogViewer / Help / Debug / Terminal …) report false, matching the pointer handler; add explicit branches when a mode gains real scrolling:

```diff
   # Handle text editing modes.
   if e.state.mode in {
     ...
   }:
     ...
     return
-
-  result.handled = true
-  result.region = e.scrollRegion(e.activeWindow)
```

fox0430 (approved):
Thanks!

PR #3208, opened 2026-08-24
Integration improvements

elcritch:
Collection of changes to improve front end embedding in a GUI. 

Progress so far!

image

### DIFF
--- a/src/moepkg/editor_render_views.nim
+++ b/src/moepkg/editor_render_views.nim
@@ -95,6 +95,26 @@ proc cursorIsInViewport(e: Editor, window: EditorWindow, layout: WindowLayout): 
     ) + cursorSegment
   cursorRow in 0 ..< layout.adjustHeight
 
+proc clampDetachedViewport(e: Editor, window: EditorWindow, layout: WindowLayout) =
+  ## Preserve detached scrolling while keeping its top position valid after
+  ## buffer edits, wrap changes, or window resizing.
+  let textBuffer = window.buffer
+  if textBuffer.isNil or textBuffer.len == 0:
+    window.viewport.topLine = 0
+    window.viewport.topWrapOffset = 0
+    return
+
+  window.viewport.topLine = clamp(window.viewport.topLine, 0, textBuffer.len - 1)
+  if layout.effectiveLineWrap:
+    let rowLayout = initRowLayout(
+      textBuffer, window.wrapCountCache, true, layout.textAreaWidth, e.tabStop
+    )
+    window.viewport.topWrapOffset = clamp(
+      window.viewport.topWrapOffset, 0, rowLayout.lineRows(window.viewport.topLine) - 1
+    )
+  else:
+    window.viewport.topWrapOffset = 0
+
 proc adjustViewportForCursor(
     viewport: ViewPort,
     cursor: BufferPosition,
@@ -278,6 +298,7 @@ proc advanceLayoutForFrame*(e: Editor, buffer: Buffer, wasResized: bool) =
       if window.viewport.topLine < 0:
         window.viewport.topLine = 0
     elif window.viewport.detachedFromCursor:
+      e.clampDetachedViewport(window, layout)
       if layout.isActiveWindow:
         activeCursorInViewport = e.cursorIsInViewport(window, layout)
     else:
--- a/src/moepkg/editor_render_views.nim
+++ b/src/moepkg/editor_render_views.nim
@@ -72,6 +72,29 @@ proc wrapPosAbove(aLine, aSeg, bLine, bSeg: int): bool {.inline.} =
   ## in the wrapped layout (line-major, then wrap-segment within a line).
   aLine < bLine or (aLine == bLine and aSeg < bSeg)
 
+proc cursorIsInViewport(e: Editor, window: EditorWindow, layout: WindowLayout): bool =
+  if window.buffer.isNil or window.buffer.len == 0:
+    return
+  let
+    rowLayout = e.rowLayoutFor(
+      window.buffer,
+      window.viewport.width,
+      e.viewportOffsetFor(window),
+      window.wrapCountCache,
+    )
+    cursorLine = max(0, min(window.cursor.line, window.buffer.len - 1))
+    cursorSegment = rowLayout.cursorCell(cursorLine, window.cursor.column).wrapSeg
+  if wrapPosAbove(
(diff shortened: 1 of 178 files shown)


src/moepkg/frontend.nim
```diff
@@ -0,0 +1,31 @@
...
+## Hosts can construct and drive an `Editor`, translate native input through
+## the frontend-neutral handlers, and read `FrontendStatus` without depending
+## on Moe's terminal application entry point.
+##
+## Call `setFrontendGitStatusEnabled(true)` once when the host displays Git
+## status, then call `tick` each frame before reading `frontendStatus`.
+
+import config, editor, frontend_input, handler
+
+export config, editor, frontend_input, handler
```
elcritch: This might be a bit overkill, but it does centralize frontend imports/exports.

elcritch:
These changes enable the GUI framework to provide "smooth" scrolling. It also let's it own the bottom status bar so it can use a common area.

In another PR I'd like to add more info for the GUI to turn off popup menus so it can handle those as well. That may expand on the "frontend subscription" setup added here.

src/moepkg/handler.nim
fox0430, line 1258: Middle-clicking while the viewport is wheel-scrolled away **rewrites the buffer with no trace on screen**. The text lands at the offscreen cursor, `advanceLayoutForFrame` takes the detached branch (`editor_render_views.nim:280`) so the viewport does not follow, and `cursorIsInViewport` returns false so the cursor is hidden as well. The only hint is the `[+]` on the tab.

**:1258 is outside this PR** (the `handler.nim` changes stop at line 810). This is not an oversight in the diff but a regression caused by introducing new state that untouched code does not handle.

`detachedFromCursor` does not exist at the merge base; this PR adds it. The previous wheel handler moved `window.cursor` along with the viewport, so the cursor was always on screen and middle-click inserted where the user could see it. Making the wheel move the viewport only is what breaks the untouched middle-click path.

Three of the paths that need the same handling are covered by this PR - `handlePaste` (`:806-810`) and both pointer-click paths (`:686` / `:733`). Only middle-click is left on `prepareForInput(false)`, which does not clear the flag because `isKeyInput = false`. Neither `middleClickPaste` (`:369`) nor `handlePasteText` (`:264`) touches it.

60-line file, PRIMARY set to `PASTED`. Seven wheel-downs, then middle-click:

```
 t.txt[+]
  22 line022                          <- screen does not move
 INSERT  ...  1/60 7/13               <- the insert did happen
flag=0                                <- cursor hidden
file after :w = PASTEDline001         <- landed on line 1, offscreen
```

Sending a bracketed paste under the same conditions goes through `handlePaste`: it **rejects** the insert with `Paste not supported in this mode`, yet still snaps back to line 1 and restores the cursor. The paste that inserts nothing re-attaches; the middle-click that actually edits does not.

### Fix

```diff
 if event.mouse.button == mouse_logic.MouseButton.Middle and
     event.mouse.kind == celina.MouseEventKind.Press:
   e.prepareForInput(false)
+  if e.windowManager.windows.len > 0:
+    e.activeWindow.viewport.detachedFromCursor = false
   e.middleClickPaste()
   return true
```

src/moepkg/frontend.nim
fox0430: It looks nice. No problem.

fox0430 (changes requested):

src/moepkg/frontend.nim
fox0430: Please export individual symbols instead.

```nim
export config.EditorConfig, config.newEditorConfig
export editor.Editor, editor.newEditor
export editor_frame.tick
export
  editor_display.FrontendStatus, editor_display.ActiveGitStatus,
  editor_display.frontendStatus, editor_display.frontendGitStatusEnabled,
  editor_display.setFrontendGitStatusEnabled
export
  editor_buffers.OpenBufferInfo, editor_buffers.activeWindowBuffers,
  editor_buffers.activateBuffer, editor_buffers.closeBuffer,
  editor_buffers.moveBuffer, editor_buffers.deleteCurrentBuffer
export
  frontend_input.GridRegion, frontend_input.PointerButton,
  frontend_input.PointerAction, frontend_input.PointerInput,
  frontend_input.ScrollInput, frontend_input.ScrollOutcome
export handler.handleTextInput, handler.handleEvent
```

This clarifies the APIs used by the frontend and prevents breaking changes.

fox0430 (changes requested):

src/moepkg/frontend.nim
elcritch: Good call, that's nicer.

src/moepkg/editor_render_views.nim
```diff
@@ -253,6 +277,9 @@ proc advanceLayoutForFrame*(e: Editor, buffer: Buffer, wasResized: bool) =
         window.viewport.topLine = selected - visible + 1
       if window.viewport.topLine < 0:
         window.viewport.topLine = 0
+    elif window.viewport.detachedFromCursor:
```
fox0430, line 300: `detachedFromCursor` is only ever cleared for the active window, so a background window that has been detached never gets its viewport re-clamped again.

`scrollTextViewport` (`handler.nim:569`) sets the flag on whichever window `scrollTargetIndex` returns, which can be a background window (terminal wheel events take the same path via `handler.nim:854-857`). The five clear sites (`handler.nim:686/733/798/810/1262`) all target `e.activeWindow`, so nothing clears it for a background window.

This branch then skips `adjustViewportForCursor` entirely, and before this PR that call was the only per-frame sanitisation of `topLine` / `topWrapOffset`. An out-of-range `topLine` is therefore never corrected.

Reproduced with 2 splits on 1 buffer:

```
wheel 60 rows over window 1 (inactive) -> w1.topLine=60, detached=true
delete 95 lines in window 0, j x6, 6 frames -> buffer len=6, w1.topLine=60, detached=true
```

Window 1 stays blank until the user clicks into it.

Could you clamp `topLine` against the buffer length and `topWrapOffset` against `topLine`'s wrap count here? Note that `adjustViewportForCursor` lines 124-131 only re-clamp `topWrapOffset`; `topLine` is currently repaired implicitly by the cursor-follow at lines 138-141, so the `topLine` clamp has to be new code. The same clamp also covers the resize case that leaves `topWrapOffset` out of range.

fox0430 (changes requested):

fox0430 (approved):
Thanks!

PR #3217, opened 2026-08-26
Add Ctrl-E and Ctrl-Y line scrolling shortcuts

elcritch:
## Summary

- add Normal-mode Ctrl-E and Ctrl-Y bindings for scrolling the viewport down/up by screen rows
- preserve the cursor while visible and move it only when it leaves the viewport
- use Moe’s shared row layout so wrapped lines and collapsed folds scroll consistently with rendering
- support numeric counts and document both shortcuts

## Testing

- exact nph v0.7.0 formatting check
- command registry, key binding, binding drift, and help synchronization tests
- focused scrolling, count, collapsed-fold, and line-wrap regression tests
- debug and release builds

### DIFF
--- a/src/moepkg/command_registry/motion_scroll.nim
+++ b/src/moepkg/command_registry/motion_scroll.nim
@@ -23,7 +23,7 @@ import std/options
 
 import pkg/results
 
-import ../[types, motion, modes, render_utils]
+import ../[types, motion, modes, render_utils, visible_rows]
 import ../buffer/[core, fold]
 
 import core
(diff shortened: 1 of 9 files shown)


src/moepkg/command_registry/motion_scroll.nim
```diff
@@ -118,6 +118,46 @@ proc handleScrollCursorBottom*(
...
+  max(1, ctx.motionController.viewportManager.viewport.height - reservedLines)
+
+proc handleScrollLines(
+    ctx: CommandContext, args: seq[string], direction: int
+): Result[(), string] =
+  let
+    count = parseCount(args)
+    visibleHeight = ctx.scrollVisibleHeight()
+    viewport = ctx.motionController.viewportManager.viewport
+    maxTopLine = max(0, ctx.buffer.len - visibleHeight)
```
fox0430: `maxTopLine = buffer.len - visibleHeight` assumes one screen row per buffer line, so under `lineWrap` the real top legitimately sits far below it.
After `G` on a 10-line buffer that wraps 4 rows per line (`visibleHeight = 6`) the authority parks the top at `(line 8, seg 2)` while `maxTopLine` is `4`: Ctrl-E clamps it *backwards* to `(4, 0)`, `adjustViewportForCursor` restores `(8, 2)` in the same frame, and the key is dead at the end of a wrapped file.

src/moepkg/command_registry/motion_scroll.nim
```diff
@@ -118,6 +118,46 @@ proc handleScrollCursorBottom*(
...
+    targetTopLine = clamp(viewport.topLine + direction * count, 0, maxTopLine)
```
fox0430: `topLine ± count` steps by buffer lines, so one press moves 1..N rows under `lineWrap`, and with folds it walks the top into a collapsed fold's interior - a position the renderer cannot express (`visibleLines` drops the marker and restarts at `fold.endLine + 1`, or draws an empty screen when the fold reaches the end of the buffer).

src/moepkg/command_registry/motion_scroll.nim
```diff
@@ -118,6 +118,46 @@ proc handleScrollCursorBottom*(
...
+    targetCursorLine = clamp(
```
fox0430: The cursor clamp pins a visible cursor into a collapsed fold

100 lines, fold `2..90` collapsed, `visibleHeight = 5`, `topLine = 0`; the screen shows `0, 1, [fold], 91, 92` with the cursor on line 92. One Ctrl-E gives `targetTopLine = 1` and upper bound `min(99, 5) = 5`, so `targetCursorLine = clamp(92, 1, 5) = 5` - fold interior -and `clampPosition` (`motion.nim:945-948`) pins it to the fold's `startLine`: **the cursor jumps 92 → 2, column 0**.

From the second press on it stalls: the top walks into the fold interior, the cursor is pinned back to `startLine`, and `adjustViewportForCursor` pulls the top back down to it.
A collapsed fold is a hard wall — Ctrl-E can never get past one. Vim scrolls the whole fold in a single press.

under `lineWrap` the same clamp leaves the cursor *below* the last row and the render scrolls straight back. With the cursor on the last row Ctrl-Y is an exact no-op: `(5, 0)` -> Ctrl-Y -> `(4, 0)` -> render -> `(5, 0)`.

```nim
test "Ctrl-E must not drag a visible cursor into a collapsed fold":
  var lines: seq[string] = @[]
  for i in 0 ..< 100: lines.add($i)
  let buffer = newTextBuffer(lines.join("\n"))
  check buffer.foldState.addFold(2, 90, collapsed = true)

  let ctx = createTestContext(buffer)
  ctx.cursor = BufferPosition(line: 92, column: 0)   # on screen
  ctx.motionController.viewportManager.viewport.topLine = 0
  ctx.motionController.viewportManager.viewport.height = 5
  ctx.state.windowDisplay.viewportReservedLines = 0
  let registry = createTestRegistry()

  check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
  check ctx.cursor.line == 92
```

src/moepkg/command_registry/motion_scroll.nim
fox0430: `maxTopLine = buffer.len - visibleHeight` assumes one screen row per buffer line, so under `lineWrap` the real top legitimately sits far below it. After `G` on a 10-line buffer that wraps 4 rows per line (`visibleHeight = 6`) the authority parks the top at `(line 8, seg 2)` while `maxTopLine` is `4`: Ctrl-E clamps it *backwards* to `(4, 0)`, `adjustViewportForCursor` restores `(8, 2)` in the same frame, and the key is dead at the end of a wrapped file.

src/moepkg/command_registry/motion_scroll.nim
```diff
@@ -118,6 +118,46 @@ proc handleScrollCursorBottom*(
...
+    if ctx.state.windowDisplay.viewportReservedLines >= 0:
+      ctx.state.windowDisplay.viewportReservedLines
+    else:
+      steadyBottomAreaHeight()
+  max(1, ctx.motionController.viewportManager.viewport.height - reservedLines)
+
+proc handleScrollLines(
+    ctx: CommandContext, args: seq[string], direction: int
+): Result[(), string] =
+  let
```
fox0430, line 199: `handleScrollLines` computes geometry in buffer lines (`topLine ± count`, buffer.len  - visibleHeight`).
The authority, `adjustViewportForCursor` (`editor_render_views.nim:75`), works in screen rows, is fold- and wrap-aware, and only rewrites `topLine` when the cursor leaves the window — so the damage propagates through the cursor. `lineWrap` defaults to `true` (`config.nim:95`).

**The cursor is dragged into a collapsed fold** (`:138-149`). With 100 lines, fold `2..90` collapsed, `visibleHeight = 5` and `topLine = 0`, the screen shows `0, 1, [fold], 91, 92`, so the cursor on line 92 is visible and Ctrl-E must not move it. But `targetCursorLine = clamp(92, 1, 5) = 5` is fold interior, and `clampPosition` (`motion.nim:945-948`) pins it to the fold's
`startLine`: the cursor jumps 92 → 2, column 0.

**Ctrl-E can never get past a collapsed fold** (`:137`). Stepping by buffer lines makes the top point at fold interior. `visibleLines` can express that top (`visible_rows.nim:91-95,118-119` - it starts below the marker); what stalls is the cursor, pinned to line 2 and rejected by `clampPosition` on every press while the next frame's authority pulls the top back to line 2. Every further press leaves `topLine=2 cursor=2`. Vim treats a collapsed fold as one screen row and steps over it in a single press.

**Under wrap the clamp is wrong and the view jumps upwards** (`:136,144`). `maxTopLine = buffer.len - visibleHeight` sits far below the real maximum top under wrap. Ten 30-column lines at width 10 wrap to three rows each, so with `visibleHeight = 6` and the cursor on the last line the legitimate top is `(7, 1)` - what `walkBackRows` returns. Press Ctrl-E there and `clamp(8, 0, 4) = 4`: instead of one row down, the view flies ten rows *up*. Even with the top at or below `maxTopLine`, `resetViewportTop` clears `topWrapOffset` by contract (`types.nim:1086-1094`), so `(4, 2)` → `(4, 0)` moves the view up two rows.

`zt`/`zz`/`zb` in the same file share the buffer-line arithmetic, and they are not blockers (`:74-119`): they never move the cursor and anchor the top to the cursor's own line, so the next frame's authority absorbs the error. These two commands are the only ones that clamp against a maximum *and* move the cursor — that is where the damage comes from. Credit where due: taking the reserve from `state.windowDisplay.viewportReservedLines` (the same steady reserve the authority fuses) is more correct than the `steadyBottomAreaHeight()` of `zt`/`zz`/`zb`.

### Fix

Move the top by N screen rows via `rowAt` / `walkBackRows` (`visible_rows.nim:153,168`) and derive
the visible cursor range from `visibleLines`. That makes all three unrepresentable rather than
three separate patches. The same `RowLayout` the authority uses can be built from
`ViewportManager.wrapCountCache` (see `motion.nim:1003`). The new top carries a wrap-segment
component, so write it with `restoreViewportTop` (`types.nim:1095`), not `resetViewportTop`.

tests/test_command_registry_integration.nim
```diff
@@ -2050,6 +2050,72 @@ suite "Handler - Scroll operations":
...
+    let registry = createTestRegistry()
+
+    check registry.execute(ctx, builtin(bcScrollLineDown), @["20"]).isOk
+    check ctx.motionController.viewportManager.viewport.topLine == 5
+    check ctx.cursor.line == 5
+
+    check registry.execute(ctx, builtin(bcScrollLineUp), @["20"]).isOk
+    check ctx.motionController.viewportManager.viewport.topLine == 0
+    check ctx.cursor.line == 4
+
```
fox0430, line 2118: Tests to add.

```nim
  test "scroll line down keeps a visible cursor out of a collapsed fold (Ctrl-E)":
    # Screen shows 0, 1, [fold 2..90], 91, 92 -- the cursor is already visible,
    # so scrolling by one row must not move it at all.
    var lines: seq[string] = @[]
    for i in 0 ..< 100:
      lines.add($i)
    let buffer = newTextBuffer(lines.join("\n"))
    check buffer.foldState.addFold(2, 90, collapsed = true)

    let ctx = createTestContext(buffer)
    ctx.cursor = BufferPosition(line: 92, column: 0)
    ctx.motionController.viewportManager.viewport.topLine = 0
    ctx.motionController.viewportManager.viewport.height = 7
    ctx.state.windowDisplay.viewportReservedLines = 2
    let registry = createTestRegistry()

    check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
    check ctx.cursor.line == 92

  test "scroll line down steps over a collapsed fold in one press (Ctrl-E)":
    # A collapsed fold occupies a single screen row, so three presses from the
    # top reach line 91: 0 -> 1 -> [fold] -> 91.
    var lines: seq[string] = @[]
    for i in 0 ..< 100:
      lines.add($i)
    let buffer = newTextBuffer(lines.join("\n"))
    check buffer.foldState.addFold(2, 90, collapsed = true)

    let ctx = createTestContext(buffer)
    ctx.cursor = BufferPosition(line: 92, column: 0)
    let viewport = ctx.motionController.viewportManager.viewport
    viewport.topLine = 0
    viewport.height = 7
    ctx.state.windowDisplay.viewportReservedLines = 2
    let registry = createTestRegistry()

    check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
    check viewport.topLine == 1
    check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
    check viewport.topLine == 2
    check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
    check viewport.topLine == 91

  test "scroll line down never moves the top backwards under lineWrap (Ctrl-E)":
    # 10 lines of 30 columns at width 10 wrap to 3 rows each, so the top may
    # legitimately sit below `buffer.len - visibleHeight` on a wrap segment.
    var lines: seq[string] = @[]
    for i in 0 ..< 10:
      lines.add("x".repeat(30))
    let buffer = newTextBuffer(lines.join("\n"))

    let ctx = createTestContext(buffer)
    ctx.state.config.standard.lineWrap = true
    ctx.cursor = BufferPosition(line: 9, column: 0)
    let viewport = ctx.motionController.viewportManager.viewport
    viewport.topLine = 4
    viewport.topWrapOffset = 2
    viewport.width = 10
    viewport.height = 8
    ctx.state.windowDisplay.viewportReservedLines = 2
    let registry = createTestRegistry()

    check registry.execute(ctx, builtin(bcScrollLineDown)).isOk
    check (
      viewport.topLine > 4 or (viewport.topLine == 4 and viewport.topWrapOffset >= 2)
    )
```

fox0430 (changes requested):

elcritch:
@fox0430 looks like there's an issue with Nim devel breaking the CI?

fox0430:
@elcritch 

> looks like there's an issue with Nim devel breaking the CI?

It seems the workflow is broken due to changes in Nim devel...

fox0430:
@elcritch 

I think this is a temporary issue, but just to be safe, I have opened an [PR](https://github.com/jiro4989/setup-nim-action/pull/528) in the workflow.
Well, the unit tests are passing, so there shouldn't be any issues with this PR.

fox0430 (approved):
Thanks

PR #3219, opened 2026-08-28
Add :x and :xit commands with write-if-modified semantics

titanomachy:
## Summary
Using :x is intentional: it writes the file only when the buffer changed. If nothing changed, it exits without touching the file, _preserving its mtime and avoiding false rebuilds or file-watcher events._

- add Vim-compatible `:x` and `:xit` commands
- write the current buffer only when it is modified, then use the existing save-and-quit lifecycle
- keep `:wq` unconditional and unchanged
- expose the new commands through parsing, completion, runtime key mapping, TUI help, and generated Markdown documentation

## Behavior

| Command | Clean buffer | Modified buffer |
| --- | --- | --- |
| `:x` / `:xit` | quit without writing | write, then quit |
| `:x!` | quit without writing | force write, then quit |
| `:wq` | write, then quit | write, then quit |

Optional filenames and the force flag are preserved when `:x` does enter the existing save-and-quit path.

## Implementation

This introduces a dedicated semantic `claSaveIfModifiedAndQuit` command-line action, while keeping `x` and `xit` as its external aliases. The command handler checks `TextBuffer.isModified`: clean buffers return through the existing application-exit result, while modified buffers reuse `executeSaveAndQuit`. This deliberately does not change the current window/application lifecycle or the behavior of the all-buffer commands.

## Testing

- targeted parser, configuration, command-table, handler, editor integration, completion, runtime key mapping, help, and documentation-sync tests pass
- release build compiles successfully
- `nph ./` passes
- full runner: 222/223 test files pass; the remaining existing flaky `tests/test_terminal_handler.nim` PTY shell-exit timing test failed on rerun and does not touch changed code

### DIFF
--- a/src/moepkg/command_line_commands.nim
+++ b/src/moepkg/command_line_commands.nim
@@ -208,6 +208,31 @@ const CommandLineCommandTable*: seq[CommandLineCommandSpec] = @[
     isCanonicalLong: false,
   ),
   CommandLineCommandSpec(
+    name: "x",
+    completionDescription: "Write if modified and quit",
+    helpEntries:
+      @[HelpEntry(syntax: ":x or :xit", description: "Write if modified and quit")],
+    action: some(claSaveIfModifiedAndQuit),
+    isCanonicalLong: false,
+    keymapBaseDescription: "Save if modified and quit",
+  ),
+  CommandLineCommandSpec(
+    name: "xit",
+    completionDescription: "Write if modified and quit",
+    helpEntries: @[],
+    action: some(claSaveIfModifiedAndQuit),
+    isCanonicalLong: true,
+    keymapBaseDescription: "Save if modified and quit",
+  ),
+  CommandLineCommandSpec(
+    name: "x!",
+    completionDescription: "",
+    helpEntries:
+      @[HelpEntry(syntax: ":x!", description: "Force write if modified and quit")],
+    action: none(CommandLineAction),
+    isCanonicalLong: false,
+  ),
+  CommandLineCommandSpec(
     name: "wqa",
     completionDescription: "Write and quit all windows",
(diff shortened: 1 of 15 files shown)


src/moepkg/command_line_commands.nim
```diff
@@ -207,6 +207,31 @@ const CommandLineCommandTable*: seq[CommandLineCommandSpec] = @[
     action: none(CommandLineAction),
     isCanonicalLong: false,
   ),
+  CommandLineCommandSpec(
+    name: "x",
```
fox0430, line 211: This new built-in collides with the shipped sample config. `example/moerc.toml:614` (and `documents/configfile.md:1225`) uses `x = { command = "quit" }` as its `[CommandAliases]` example - added in #2482, after #2413 had freed the name. User aliases are applied on top of the built-ins, so uncommenting the shipped example silently turns `:x` back into `:q`. Please switch the example to a name that isn't a built-in.

### Examples

example/moerc.toml: 611,  614
```diff
-# Usage: :x => quit, :ww => saveall
+# Usage: :qq => quit, :ww => saveall
# Built-in aliases (from loadDefaultConfig) can be overridden.

-# x = { command = "quit" }
+# qq = { command = "quit" }
ww = { command = "saveall", description = "Save all buffers" }
```
documents/configfile.md: 1225
```diff
[CommandAliases]
-x = { command = "quit" }
+qq = { command = "quit" }
ww = { command = "saveall", description = "Save all buffers" }
```

fox0430 (changes requested):

fox0430 (approved):
Thanks!

src/moepkg/command_line_commands.nim
titanomachy: done!

PR #3220, opened 2026-08-30
feat: add GUI pointer based text selection

elcritch:
Adds APIs for text selection using a mouse pointer.

### DIFF
--- a/src/moepkg/handler.nim
+++ b/src/moepkg/handler.nim
@@ -637,9 +637,8 @@ proc pointerTabPress(e: Editor, input: PointerInput): bool =
         buffersToShow.add(bufOpt.get)
     if buffersToShow.len == 0:
       buffersToShow = @[window.buffer]
-    let tabIdx = hitTestTabLine(
-      buffersToShow, window.mode, 0, window.viewport.width, input.column
-    )
+    let tabIdx =
+      hitTestTabLine(buffersToShow, window.mode, 0, window.viewport.width, input.column)
     if tabIdx >= 0:
       e.switchToWindowBuffer(tabIdx)
       return true
@@ -732,7 +731,9 @@ func pointerGranularity(
     psgCharacter
 
 proc pointerRange(
-    buffer: TextBuffer, position: BufferPosition, granularity: PointerSelectionGranularity
+    buffer: TextBuffer,
+    position: BufferPosition,
+    granularity: PointerSelectionGranularity,
 ): tuple[first, last: BufferPosition] =
   case granularity
   of psgWord:
@@ -791,10 +792,7 @@ proc updatePointerSelection(e: Editor, position: BufferPosition) =
     e.state.pointerSelection.selectionStarted = true
 
   e.state.visualSelection = VisualSelection(
-    start: anchor,
-    current: focus,
-    active: true,
-    kind: gesture.granularity.visualKind,
+    start: anchor, current: focus, active: true, kind: gesture.granularity.visualKind
   )
   e.cursor = focus
   e.activeWindow.viewport.detachedFromCursor = false
@@ -808,9 +806,7 @@ proc collapseVisualSelectionForPointer(e: Editor) =
     else:
       e.setMode(returnMode)
 
-proc beginPointerSelection(
-    e: Editor, input: PointerInput, hit: PointerTextHit
-) =
+proc beginPointerSelection(e: Editor, input: PointerInput, hit: PointerTextHit) =
   e.activatePointerWindow(hit.windowIndex)
 
   let
@@ -850,15 +846,16 @@ proc updatePointerGesture(e: Editor, input: PointerInput): bool =
     return false
 
   let gesture = e.state.pointerSelection
-  if gesture.windowIndex != e.windowManager.activeWindowIndex or
-      gesture.windowIndex < 0 or gesture.windowIndex >= e.windowManager.windows.len or
+  if gesture.windowIndex != e.windowManager.activeWindowIndex or gesture.windowIndex < 0 or
+      gesture.windowIndex >= e.windowManager.windows.len or
       e.activeBuffer.id != gesture.bufferId:
     e.state.pointerSelection.active = false
     return false
 
   let position = e.pointerPositionInWindow(input, gesture.windowIndex)
   if position.isSome:
-    let moved = position.get != gesture.anchorFirst or gesture.anchorFirst != gesture.anchorLast
+    let moved =
+      position.get != gesture.anchorFirst or gesture.anchorFirst != gesture.anchorLast
     if gesture.selectionStarted or input.action == paDrag or moved:
       e.updatePointerSelection(position.get)
   true
--- a/src/moepkg/handler.nim
+++ b/src/moepkg/handler.nim
@@ -660,6 +660,7 @@ proc pointerTabPress(e: Editor, input: PointerInput): bool =
         e.activatePointerWindow(i)
         e.switchToWindowBuffer(tabIdx)
         return true
+  return false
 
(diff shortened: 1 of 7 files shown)


src/moepkg/handler.nim
```diff
@@ -608,130 +608,285 @@ proc handleScrollInputCore(e: Editor, input: ScrollInput): ScrollOutcome =
...
+      if oldSelection.active: oldSelection.start else: oldCursor
+    else:
+      pressedRange.first
+
+  e.state.pointerSelection = PointerSelectionGesture(
+    active: true,
+    windowIndex: hit.windowIndex,
+    bufferId: e.activeBuffer.id,
+    anchorFirst: anchor,
+    anchorLast: if extendSelection: anchor else: pressedRange.last,
```
fox0430: With shift modifier, `anchorLast` collapses to a single point (`= anchor`). In `updatePointerSelection`'s `psgWord` branch, `target.first <= anchorLast` spuriously matches and `focus` collapses to `anchorLast`. Shift+double-clicking a word that overlaps the anchor produces an empty selection (e.g. cursor at (0,8) inside "beta", shift+double-click at (0,7) -> start=current=(0,8)).

Preserve `pressedRange.last` for word/line granularity even under shift-extend.
```nim
    anchorLast:
      if extendSelection and granularity == psgCharacter: anchor
      else: pressedRange.last,
```

Add a case in `tests/test_handler.nim` verifying that shift+double-click on the anchor word selects the whole word.

src/moepkg/handler.nim
```diff
@@ -608,130 +608,285 @@ proc handleScrollInputCore(e: Editor, input: ScrollInput): ScrollOutcome =
...
+func visualMode(granularity: PointerSelectionGranularity): EditorMode =
+  case granularity
+  of psgLine: EditorMode.VisualLine
+  of psgBlock: EditorMode.VisualBlock
+  of psgCharacter, psgWord: EditorMode.Visual
+
+proc enterPointerVisualMode(e: Editor, granularity: PointerSelectionGranularity) =
+  if not e.state.mode.isVisualAllMode:
+    let previousMode = e.state.mode
+    if previousMode in {EditorMode.Insert, EditorMode.Replace}:
```
fox0430: While `state.mode == Normal` and `state.insertNormalMode == true` (Ctrl-O window), this gate evaluates false and `finalizeCurrentWindowForMouseJump` is never called. `finalize` itself handles `insertNormalMode` explicitly at :488, but this call site's gate is limited to Insert/Replace and misses the case. The Insert transaction stays open across Visual, so subsequent Insert edits get grouped into a single undo unit.

```nim
    if previousMode in {EditorMode.Insert, EditorMode.Replace} or e.state.insertNormalMode:
      e.finalizeCurrentWindowForMouseJump()
      e.state.previousMode = EditorMode.Normal
```

Add a unit test that enters pointer Visual from `insertNormalMode=true` and asserts the transaction is closed on exit.

src/moepkg/handler.nim
```diff
@@ -608,130 +608,285 @@ proc handleScrollInputCore(e: Editor, input: ScrollInput): ScrollOutcome =
...
+  if windowIndex < 0 or windowIndex >= e.windowManager.windows.len:
+    return none(BufferPosition)
+
+  let
+    window = e.windowManager.windows[windowIndex]
+    vp = window.viewport
+  if input.column < vp.x or input.column >= vp.x + vp.width:
+    return none(BufferPosition)
+
+  if e.windowManager.windows.len > 1:
```
fox0430, line 677: The multi-window branch passes `input.row` to `screenToBufferPosition` unadjusted. Each window renders its tab line at `viewport.y` (`editor_render_views.nim:387`), so a click on the tab-line row that misses every tab (`pointerTabPress` at :647-662 falls through) is interpreted as content — `screenY=0` jumps the cursor to (0,0). The single-window branch at :697 subtracts `tabLineOffset` and is inconsistent with this branch.

Subtract `tabLineOffset` in the multi-window branch too. Also add an explicit `return false` at `pointerTabPress` :662.
```nim
  if e.windowManager.windows.len > 1:
    if input.row < vp.y or input.row >= vp.y + vp.height:
      return none(BufferPosition)
    let
      maxBottomY = findMaxBottomY(e.windowManager.windows)
      reservedLines = e.steadyReservedLines(vp.y + vp.height == maxBottomY)
      tabLineOffset = if e.showTabLine: TabLineHeight else: 0
      adjustedRow = input.row - tabLineOffset
    return screenToBufferPosition(
      vp, window.buffer, input.column, adjustedRow,
      e.gutterWidth(window), reservedLines, e.lineWrap,
      e.tabStop, e.wrapWidth(window), window.wrapCountCache,
    )
```

Add a case in `tests/test_handler.nim`: with multiple windows and `showTabLine=true`, a click on the empty tab-line area must not move the cursor.

src/moepkg/handler.nim
```diff
@@ -608,130 +608,285 @@ proc handleScrollInputCore(e: Editor, input: ScrollInput): ScrollOutcome =
...
+
+proc updatePointerSelection(e: Editor, position: BufferPosition) =
+  let
+    gesture = e.state.pointerSelection
+    target = e.activeBuffer.pointerRange(position, gesture.granularity)
+  var
+    anchor = gesture.anchorFirst
+    focus = target.last
+
+  if gesture.granularity == psgWord:
```
fox0430, line 785: The `psgWord` branch is drag-shaped and assumes `anchorFirst/anchorLast` bracket a word range; under shift-extend both collapse to the single-point `anchor`, so `elif target.first <= anchorLast: focus = anchorLast` empties the selection (cursor at (0,8) inside "beta", shift+dblclick at (0,7) → start=current=(0,8)). Setting `anchorLast = pressedRange.last` regresses cursor-after-word cases: `target.last < anchorFirst` fires, `anchor` jumps to the word's end, and cursor-to-word extend disappears.

Add `extendMode: bool` to `PointerSelectionGesture` and split the psgWord branch (an `anchorFirst==anchorLast` heuristic misfires on single-char words).
```nim
if gesture.granularity == psgWord:
  if gesture.extendMode:
    if target.first <= gesture.anchorFirst and
        gesture.anchorFirst <= target.last:
      anchor = target.first; focus = target.last
    elif target.first >= gesture.anchorFirst:
      focus = target.last 
    else:
      focus = target.first
  else:
    if target.last < gesture.anchorFirst:
      anchor = gesture.anchorLast; focus = target.first
    elif target.first <= gesture.anchorLast:
      focus = gesture.anchorLast
```

Tests: 3 cases: 1. anchor-inside-word , 2.  whole word, word before/after cursor 3. extend to cursor.
A single case cannot catch the cursor-after-word regression.

fox0430 (changes requested):

src/moepkg/handler.nim
elcritch: Fixed in cd9ea128. Pointer Visual entry now finalizes Insert-normal sessions as well as Insert/Replace, and the regression verifies the transaction and insert-normal state are cleared.

src/moepkg/handler.nim
elcritch: Fixed in cd9ea128. Split-window hit testing now subtracts the tab-line offset, pointerTabPress returns false explicitly on a miss, and the regression verifies an empty tab-line click does not move the cursor.

src/moepkg/handler.nim
elcritch: Fixed in cd9ea128. PointerSelectionGesture now records extendMode and word selection has separate Shift-extension logic. Regressions cover the caret inside, before, and after the target word.

fox0430 (approved):
Thanks!

PR #3230, opened 2026-08-30
Add forced Insert mode option

elcritch:
## Summary

- add Standard.forceInsertMode for GUI-style editing without Normal-mode commands
- keep editable windows in a valid Insert session across Escape, Ctrl-C, replay, config reload, and window changes
- reuse Moe's Insert transaction lifecycle and preserve Visual and special modes
- document the option and cover config, frontend, input, replay, and mouse behavior

## Tests

- atlas-run build
- 218/218 eligible Atlas test programs passed (known macOS/environment-dependent cases excluded)
- focused config, frontend, handler, config-reload, runtime-keymapping, macro-replay, and docs-sync tests passed
- nph check passed for changed formatter-clean Nim files

### DIFF
--- a/src/moepkg/command_handlers/mode_dispatchers.nim
+++ b/src/moepkg/command_handlers/mode_dispatchers.nim
@@ -159,7 +159,26 @@ proc replayCountedInsert(buffer: TextBuffer, state: EditorState) =
     cursor = inserted.get.cursor
   state.cursor = cursor
 
-proc finalizeInsertExit(
+proc beginInsertModeSession*(
+    buffer: TextBuffer, state: EditorState, replayCount = 0, replayLineEntry = false
+): Result[void, string] =
+  ## Begin the transaction and tracking shared by all ordinary Insert entries.
+  ## An existing transaction belongs to Insert-Normal or an entry command that
+  ## edited the buffer before returning its mode transition.
+  if not buffer.inTransaction:
+    let transactionResult =
+      buffer.beginTransaction("Insert mode edit", cursorPos = some(state.cursor))
+    if transactionResult.isErr:
+      return err(transactionResult.error)
+
+  if state.editState.insertModeStartPos.isNone:
+    state.editState.insertModeStartPos = some(state.cursor)
+    state.editState.insertReplayCount = replayCount
+    state.editState.insertReplayLineEntry = replayLineEntry
+
+  ok()
+
+proc finalizeInsertExit*(
     buffer: TextBuffer, state: EditorState
 ): Result[string, string] =
   ## Shared leaving-Insert cleanup: snippet/auto-indent teardown, `.`-repeat +
@@ -254,6 +273,15 @@ proc handleInsertMode*(
   of imrHandled:
     # Check if we're leaving Insert mode
     if r.modeTransition.isSome and r.modeTransition.get != EditorMode.Insert:
+      # Forced Insert mode disables Ctrl-o's one-shot Normal command. Keep the
+      # current transaction open so the key is a true no-op.
+      if state.config.standard.forceInsertMode and state.insertNormalMode and
+          r.modeTransition.get == EditorMode.Normal:
+        state.insertNormalMode = false
+        return HandlerResult(
+          kind: hrHandled, modeTransition: none(EditorMode), statusMessage: ""
+        )
+
       # Ctrl-o (insert-normal mode): skip transaction commit/cleanup,
       # keep insert state intact so we can resume after one Normal command
       if state.insertNormalMode:
--- a/src/moepkg/command_handlers/mode_dispatchers.nim
+++ b/src/moepkg/command_handlers/mode_dispatchers.nim
@@ -178,6 +178,50 @@ proc beginInsertModeSession*(
 
   ok()
 
+proc recordLastInsertEdit(transaction: buffer.BufferTransaction, state: EditorState) =
+  let insertedText = extractInsertedText(transaction)
+  if insertedText.len == 0 or state.editState.insertModeStartPos.isNone:
+    return
+
+  if state.editState.substituteContext.isSome:
+    let subCtx = state.editState.substituteContext.get
+    state.editState.lastEditCommand = some(
+      types.LastEditCommand(
+        kind: types.lecSubstitute,
+        substituteText: insertedText,
+        substituteCount: subCtx.deleteCount,
+        substituteKind: subCtx.kind,
+      )
+    )
+  else:
+    state.editState.lastEditCommand = some(
+      types.LastEditCommand(
+        kind: types.lecInsertText,
+        insertedText: insertedText,
+        insertPosition: state.editState.insertModeStartPos.get,
+      )
+    )
+
+proc commitInsertModeBoundary*(
+    buffer: TextBuffer, state: EditorState
+): Result[void, string] =
+  ## Commit one Insert-mode undo group without leaving Insert mode. Unlike an
+  ## Escape or Ctrl-C exit, this keeps snippet and auto-indent state active.
+  if buffer.currentTransaction.isSome:
+    recordLastInsertEdit(buffer.currentTransaction.get, state)
+
+  state.editState.insertModeStartPos = none(BufferPosition)
+  state.editState.insertReplayCount = 0
+  state.editState.insertReplayLineEntry = false
+  state.editState.substituteContext = none(types.SubstituteContext)
+
+  if buffer.inTransaction:
+    let transactionResult = buffer.commitTransaction()
+    if transactionResult.isErr:
+      return err("Failed to commit transaction: " & transactionResult.error)
+
+  ok()
+
 proc finalizeInsertExit*(
     buffer: TextBuffer, state: EditorState
 ): Result[string, string] =
@@ -221,33 +265,14 @@ proc finalizeInsertExit*(
             break
       state.editState.visualBlockInsertContext = none(types.VisualBlockInsertContext)
 
-    if insertedText.len > 0:
-      if state.editState.substituteContext.isSome:
-        let subCtx = state.editState.substituteContext.get
-        state.editState.lastEditCommand = some(
-          types.LastEditCommand(
-            kind: types.lecSubstitute,
-            substituteText: insertedText,
-            substituteCount: subCtx.deleteCount,
-            substituteKind: subCtx.kind,
-          )
-        )
-      else:
-        state.editState.lastEditCommand = some(
-          types.LastEditCommand(
-            kind: types.lecInsertText,
-            insertedText: insertedText,
-            insertPosition: state.editState.insertModeStartPos.get,
-          )
-        )
-
+    recordLastInsertEdit(transaction, state)
     # Replay before commit so [count]i repeats share the same undo group.
     replayCountedInsert(buffer, state)
 
-    state.editState.insertModeStartPos = none(BufferPosition)
-    state.editState.insertReplayCount = 0
-    state.editState.insertReplayLineEntry = false
-    state.editState.substituteContext = none(types.SubstituteContext)
+  state.editState.insertModeStartPos = none(BufferPosition)
+  state.editState.insertReplayCount = 0
+  state.editState.insertReplayLineEntry = false
+  state.editState.substituteContext = none(types.SubstituteContext)
 
   if buffer.inTransaction:
     let transactionResult = buffer.commitTransaction()
@@ -262,6 +287,17 @@ proc finalizeInsertExit*(
 
   ok("")
 
+proc finalizeInsertInterrupt*(
+    buffer: TextBuffer, state: EditorState
+): Result[void, string] =
+  ## Finish Insert mode for Ctrl-C without Escape-only count replay or visual
+  ## block replication. The typed text is still recorded for dot-repeat.
+  state.snippetSession.active = false
+  clearAutoIndentIfUnedited(buffer, state)
+
+  state.editState.visualBlockInsertContext = none(types.VisualBlockInsertContext)
+  commitInsertModeBoundary(buffer, state)
+
 proc handleInsertMode*(
     manager: HandlerManager, editor: Editor, keyCombo: KeyCombo
 ): HandlerResult =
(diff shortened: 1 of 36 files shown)


src/moepkg/handler.nim
```diff
@@ -1076,25 +1083,32 @@ proc handleInterruptCore(e: Editor): bool =
     # Other file edit modes (Insert, Visual, Replace, etc.): switch to Normal mode
     let activeBuffer = e.activeBuffer()
 
-    # Commit transaction when leaving Insert or Replace mode
-    if e.state.mode in {EditorMode.Insert, EditorMode.Replace}:
+    # Finalize the complete Insert session so replay/snippet/block state cannot
+    # leak across Ctrl-C. Replace mode only owns a transaction and history.
+    if e.state.mode == EditorMode.Insert:
+      let finalizeResult = finalizeInsertExit(activeBuffer, e.state)
```
fox0430: Ctrl-C now runs the whole `finalizeInsertExit` body, and 2 of the steps in there belong to Escape alone.

`replayCountedInsert` repeats the count: `3i` `x` `<C-c>` gives `xxx` where develop and Vim give `x` (Vim on a terminal only — under `vim -s` Ctrl-C is not an interrupt and gives `xxx`). The visual-block replication (mode_dispatchers.nim:203-222) replays the typed text into every line of `visualBlockInsertContext`: `<C-v>` `j` `j` `I` `X` `<C-c>` on `aaa`/`bbb`/`ccc` gives `Xaaa`/`Xbbb`/`Xccc` where develop and Vim give `Xaaa`/`bbb`/`ccc`.

Resetting `insertReplayCount` covers neither the block context nor the second Ctrl-C route: the KeyCombo C-c (key_bindings/insert_bindings.nim:29, sharing Escape's command) already repeats the count on develop, so give it its own Insert-mode command or route it into `handleInterruptCore`. develop is not clean on the block context either — it leaves `visualBlockInsertContext` set after Ctrl-C, so the landing point there is "clear the context without replicating", not a revert.

Keep the rest - the same routing made Ctrl-C set `lastEditCommand`, which matches Vim.

src/moepkg/handler.nim
```diff
@@ -1076,25 +1083,32 @@ proc handleInterruptCore(e: Editor): bool =
     # Other file edit modes (Insert, Visual, Replace, etc.): switch to Normal mode
     let activeBuffer = e.activeBuffer()
 
-    # Commit transaction when leaving Insert or Replace mode
-    if e.state.mode in {EditorMode.Insert, EditorMode.Replace}:
+    # Finalize the complete Insert session so replay/snippet/block state cannot
+    # leak across Ctrl-C. Replace mode only owns a transaction and history.
+    if e.state.mode == EditorMode.Insert:
```
fox0430, line 1089: src/moepkg/handler.nim:1088 and :1101

`substituteContext` now survives Ctrl-C, so the next `i`...`<Esc>` records `lecSubstitute` and `.` deletes text instead of repeating the insert: `cc` `N` `<C-c>` then `i` `Q` `<Esc>` `.` on `QN` gives `Q`, not `QQN`. Replace at :1101 loses the same reset (`cc` `N` `<Esc>` `R` `y` `<C-c>`).

The reset sits behind `insertModeStartPos.isSome` (mode_dispatchers.nim:199, :250) and `s`/`S`/`cc` assign `state.mode` directly (command_registry/edit.nim:542, :621), so `beginInsertModeSession` never sets that field for them. Restoring the Ctrl-C resets only masks it — Escape reaches the same corruption on develop. Have the substitute commands set `insertModeStartPos` like every other Insert entry: moving the reset out of the guard instead stops the corruption but leaves `s`/`S`/`cc` recording no `.` at all (see the pre-existing section).

Test: `cc` `N` `<C-c>` leaves `substituteContext` none, then `i` `Q` `<Esc>` `.` on`QN` gives `QQN`; same pair for the Replace exit, `cc` `N` `<Esc>` `R` `y` `<C-c>`.

src/moepkg/handler.nim
```diff
@@ -1076,25 +1083,32 @@ proc handleInterruptCore(e: Editor): bool =
...
+      let finalizeResult = finalizeInsertExit(activeBuffer, e.state)
+      if finalizeResult.isErr:
+        e.state.statusMessage = finalizeResult.error
```
fox0430: Insert lost its `logError` on commit failure, and `finalizeInsertExit` returns `err` unlogged (mode_dispatchers.nim:255). The message still renders for one frame, but `prepareForInput` wipes `statusMessage` on the next keystroke (handler.nim:945), so nothing durable survives. The Replace branch below still logs.

src/moepkg/editor_mode.nim
```diff
@@ -17,21 +17,33 @@
...
-    e.state.statusMessage = msg
+    requestedMode
+
+proc enforceModePolicy*(e: Editor) =
+  ## Keep an editable window in a valid Insert session when Normal mode is
+  ## disabled. Special and Visual modes remain available.
+  if not e.forceInsertMode or e.currentMode notin {EditorMode.Normal, EditorMode.Insert}:
+    return
+
+  let transactionResult = beginInsertModeSession(e.activeBuffer, e.state)
```
fox0430, line 67: the transaction has no boundary a user of this option would ever hit. Escape still commits, but nothing else does, so the whole session is one undo entry and it outlives buffer switches; with Ctrl-o disabled there is no way to reach `u` either. `trimTrailingWhitespaceOnSave` also stops running: editor_file.nim:211 skips trimming while `inTransaction`, expecting a retry on the next save that now never comes (`autoSave` included, editor_frame.nim:349). Needs an idle / newline / pre-save boundary.

Test: typing grows `undoStack` without an Escape - today five keystrokes leave it at 0 - and `trimTrailingWhitespaceOnSave` still trims on a save taken mid-session.

src/moepkg/editor_mode.nim
```diff
@@ -17,21 +17,33 @@
...
+  if e.forceInsertMode and requestedMode == EditorMode.Normal:
+    EditorMode.Insert
   else:
-    e.state.statusMessage = msg
+    requestedMode
+
+proc enforceModePolicy*(e: Editor) =
+  ## Keep an editable window in a valid Insert session when Normal mode is
+  ## disabled. Special and Visual modes remain available.
+  if not e.forceInsertMode or e.currentMode notin {EditorMode.Normal, EditorMode.Insert}:
```
fox0430: no read-only check here, and `beginTransaction` (buffer/undo.nim:162) has none either, so this bypasses the read-only refusals at normal_handler.nim:452 and command_registry/edit.nim:1115, :1153: the window shows INSERT while every keystroke is rejected with "Buffer is read-only". 

Reached via the QuickRun / buildOnSave output splits (handler.nim:1523, :1564) and `moe -R` (editor_buffers.nim:687, :695). Add `e.activeBuffer.readOnly` to the condition.

buildOnSave output splits (handler.nim:1523, :1564) and `moe -R` (moe.nim:335 for the first file, editor_buffers.nim:687, :695 for the rest). Add `e.activeBuffer.readOnly` to the condition.

Test: a read-only buffer never enters Insert - one case per entry point: `moe -R` with a single file (moe.nim:335), `moe -R` with several (editor_buffers.nim:687, :695), and the QuickRun / buildOnSave output split - plus a unit test that `beginTransaction` on a read-only buffer errors.

src/moepkg/handler.nim
```diff
@@ -943,6 +943,7 @@ proc prepareForInput(e: Editor, isKeyInput: bool) =
   # Clear status message from previous event cycle:
   # messages persist for one render frame, then clear on next input)
   e.state.statusMessage = ""
+  e.enforceModePolicy()
```
fox0430, line 947: no overlay check. `currentMode` is `e.activeWindow.mode` (types/editor_types.nim:174) and ignores overlays, while `enterCommandOverlay` (types.nim:1187) preserves the base mode, so turning the option on under an open `:` prompt (`liveReloadOfConf`) flips the base mode to Insert and opens a transaction mid-command-line.
Narrow to reach, but the guard belongs at the top of `enforceModePolicy`.

Test: `applyConfigSettings` turning the option on while a `:` overlay is open leaves the base mode untouched.

src/moepkg/editor_mode.nim
```diff
@@ -17,21 +17,33 @@
...
-    e.notifyPopup(msg, level)
+func effectiveMode*(e: Editor, requestedMode: EditorMode): EditorMode =
+  ## Resolve a requested mode through the configured editing policy.
+  if e.forceInsertMode and requestedMode == EditorMode.Normal:
+    EditorMode.Insert
   else:
-    e.state.statusMessage = msg
+    requestedMode
+
+proc enforceModePolicy*(e: Editor) =
```
fox0430, line 49: the option strands the user with no way to save or quit.** The Command overlay opens only from a Normal-mode `:`, Insert's default bindings are Escape and C-c alone (key_bindings/insert_bindings.nim:29), insert_handler.nim:1379 refuses `ctOverlaySwitch`, and this PR makes Ctrl-o a no-op (mode_dispatchers.nim:278).

Measured: `:` inserts a literal `:` and Ctrl-C stays in Insert, so `:w`/`:q`/`:qa` are unreachable and the only exit is killing the process. The lone hatch is a pre-configured `imap` onto the `exec.cmdline.*` bridge (insert_handler.nim:1394), which neither documents/configfile.md nor example/moerc.toml mentions - keep one built-in route to the command line, or document the requirement. The proc's "Visual and special modes remain available" comment is wrong for the same reason.

Test: a built-in key reaches the command line so `:w`/`:qa` are typeable from a default config; today `:` inserts a literal `:` and leaves `state.overlay` none.

fox0430 (changes requested):

src/moepkg/handler.nim
elcritch: Fixed in 058712e3. Ctrl-C now uses a separate `finalizeInsertInterrupt` path, raw frontend `KeyCombo` Ctrl-C is routed through it, counted insert replay and block replication remain Escape-only, block context is cleared, and the typed edit is still recorded. Added regressions for `3ix<C-c>` and visual-block `IX<C-c>`.

src/moepkg/handler.nim
elcritch: Fixed in 058712e3. The substitute entry commands now initialize `insertModeStartPos`; Insert and Replace Ctrl-C clear substitute state. Added the requested `cc N <C-c>; i Q <Esc>; .` and Replace-exit dot-repeat regressions.

src/moepkg/editor_mode.nim
elcritch: Fixed in 058712e3. Forced Insert sessions now commit/restart after 500 ms idle, and regular save, save-all, and auto-save commit a pre-save boundary before trimming. Boundaries also refresh Insert tracking and dot-repeat metadata. Added undo-stack, direct-save trim, and auto-save trim regressions.

src/moepkg/editor_mode.nim
elcritch: Fixed in 058712e3. The policy excludes read-only buffers and `beginTransaction` now rejects them. Policy enforcement was added for first/additional `-R` files, buffer activation, auto-splits, and QuickRun/build output splits. Tests cover first `-R`, both additional-file paths, real async QuickRun/build splits, and transaction rejection.

src/moepkg/handler.nim
elcritch: Fixed in 058712e3. `enforceModePolicy` now returns while any overlay is active. The config-reload test enables forced Insert mode under an open command overlay and verifies that the Normal base mode and transaction state remain untouched.

src/moepkg/editor_mode.nim
elcritch: Fixed in 058712e3. Ctrl-O is now the built-in forced-mode route to the Command overlay, without exposing a one-shot Normal command. The test verifies that `:w` is typeable, and the generated/config example descriptions now document Ctrl-O.

src/moepkg/handler.nim
```diff
@@ -1340,13 +1353,20 @@ proc handleKeyCombo*(e: Editor, keyCombo: KeyCombo): bool =
...
     return overlayResult.get
 
   # Handle LSP popups (CodeLens picker, Hover popup); some keys fall through
   let popupResult = e.handlePopupKeyCombo(keyCombo)
   if popupResult.isSome:
     return popupResult.get
 
+  # GUI frontends may deliver Ctrl-C as a KeyCombo rather than an interrupt
+  # event. Route both forms through the same non-Escape Insert cleanup.
+  if e.state.mode == EditorMode.Insert and not keyCombo.isSpecial and
```
fox0430: The new KeyCombo branch runs before `manager.handleKeyCombo`, so it shadows the registry rather than joining it: the combo never reaches `resolveBuiltin` (insert_handler.nim:1377) and any binding other than `switch-to-normal` is dead, `:iunmap C-c` included. An Insert `C-c` bound to `insert-newline` inserts the line and stays in Insert at `826fe1e8`; at this head it inserts nothing and leaves Insert. Nothing gates the branch on `forceInsertMode`, so this is the one finding that reaches users who never enable the feature. Entering it only when the resolved command is `switch-to-normal` keeps both the GUI route and the remap.

Test: an Insert-mode `C-c` rebound to a non-exit command runs that command instead of leaving
Insert

src/moepkg/command_handlers/mode_dispatchers.nim
```diff
@@ -254,6 +309,17 @@ proc handleInsertMode*(
   of imrHandled:
     # Check if we're leaving Insert mode
     if r.modeTransition.isSome and r.modeTransition.get != EditorMode.Insert:
+      # Forced Insert mode uses Ctrl-o as its built-in route to Command mode.
+      if state.config.standard.forceInsertMode and state.insertNormalMode and
```
fox0430, line 313: Ctrl-O now opens the Command overlay instead of one Normal command, and nothing replaces that route: Insert binds only `Escape` and `C-c` (key_bindings/insert_bindings.nim:29), there is no `:undo` ex command, and a `[Keybindings]` rebind to `edit.undo` dies at insert_handler.nim:1401 - which dispatches only `insert.backspace`, `insert.delete`, `insert.newline` and `exec.cmdline.*` - as an `imrUnhandled` that result_processor.nim:143 drops. Adding `:undo` / `:redo` as ex commands is the cheapest fix: it fills the Ctrl-O command line and the `exec.cmdline.*` bridge at once.

It needs a transaction guard first. `undo`/`redo` (buffer/undo.nim:370) check `readOnly` and an empty stack but not `inTransaction`, so on a buffer holding a committed `abc` plus an uncommitted `de`, `undo()` returns ok and leaves the line `de` - a state never typed. That code is outside this diff and `iabc<Esc>ade`, `C-o`, `u` already reproduces it on `develop`, but forced Insert would make it the common case. Committing the in-flight transaction first matches vim's `i_CTRL-O u`.

Test: undo and redo from forced Insert restore the previous typing group, and an undo with an
open transaction does not strand the in-flight text on a rewound group.

fox0430 (changes requested):

src/moepkg/handler.nim
elcritch: Fixed in 0dcc33b8. Ctrl-C interrupt cleanup now runs only when Insert-mode binding resolution returns the built-in `switch-to-normal` command; remapped Ctrl-C commands execute normally and remain in Insert mode. Added a regression test for `C-c -> insert-newline`.

src/moepkg/command_handlers/mode_dispatchers.nim
elcritch: Fixed in 0dcc33b8. Added `:undo`/`:redo` (and `:u`), commit the active Insert transaction before history operations, resume a valid Insert session afterward, and reject direct buffer undo/redo while a transaction is open. Tests cover forced `Ctrl-O :undo/:redo`, ordinary `Ctrl-O u`/`Ctrl-O Ctrl-R`, and transaction safety.

fox0430 (approved):
Thanks!

PR #3240, opened 2026-09-03
Add Windows-safe embedded frontend mode

elcritch:
## Summary

- add `-d:moe.embedded` for frontend-owned platform integration
- exclude Moe's terminal state, handler, renderer, and POSIX PTY modules from embedded builds instead of substituting fake PTYs
- route editor-core Celina imports through an embedded boundary that keeps Celina's POSIX terminal/event-I/O/app runtime out of the build
- disable Moe clipboard detection by default and guard POSIX shell-dependent runtime features in embedded Windows builds
- use Chronos' native Windows dispatcher-handle cleanup
- add a native Windows embedding smoke job around `moepkg/frontend`

The standalone terminal editor remains POSIX-only. Embedded frontends retain the editor engine and platform-neutral rendering/input value types, own their native clipboard and terminal UI, and receive an explicit unavailable status for Moe terminal commands.

## Validation

- `nim c -r -d:moe.embedded tools/embedded_smoke.nim`
- `nim c --forceBuild:on --compileOnly --os:windows --cpu:amd64 -d:moe.embedded tools/embedded_smoke.nim`
- generated Windows C verified to contain neither `termios.h` nor Celina's terminal module
- embedded compiler unit list verified no Moe terminal/PTY/ANSI or Celina terminal/event-I/O/app modules enter the build
- `nim c -r tests/test_frontend.nim`
- `nim c -r tests/test_terminal_tabs.nim`
- `nim c -r tests/test_dispatch_sub_state_mode.nim`
- `nim c --compileOnly src/moe.nim`
- native compile of `tests/test_terminal_handler.nim`, `tests/test_poll_terminal_windows.nim`, and `tests/test_lsp_worker.nim`
- native Windows CI embedding job passes

## Final status

- Final head: `d230e3faf84df5f582aa25bad05ff35005c7f070`
- `upstream/develop` remains at `646441e97ba20b9e1e5cd890e1f44a687966a7aa`; it is an ancestor of the fork branch, with 0 upstream-only commits and 7 fork-only commits.
- All 15 PR checks passed: native Windows embedded frontend, four unit-test shards, Linux/macOS build matrix, LSP end-to-end, integration, nph, and actionlint.


src/moepkg/uri_utils.nim
```diff
@@ -186,11 +189,15 @@ proc openExternalUri*(uri: string): bool =
...
   ## The process is fire-and-forget (non-blocking).
-  let cmd = when defined(macosx): "open" else: "xdg-open"
-
-  try:
-    let p = startProcess(cmd, args = @[uri], options = {poUsePath})
-    p.close()
-    return true
-  except OSError:
-    return false
+  when defined(moe.embedded):
```
fox0430: The condition at `uri_utils.nim:192` is `defined(moe.embedded)` alone, so `gf` cannot open a URL even in embedded builds on Linux/macOS. And since it just returns `false`, `file_ops.nim:138-141` renders `"Failed to open: " & uri` — a failure shown for something never attempted. The four other features this PR disables in embedded builds (terminal `editor_ops.nim:338`, clipboard `clipboard_embedded.nim:41-42`, git diff `git_diff.nim:457`, QuickRun `quick_run_utils.nim:271-273`) all return an explicit status carrying a reason; this is the one place that departs from that contract. A `bool` has no room for a reason, so either policy below means changing the return type or the call site.

src/moepkg/uri_utils.nim
```diff
@@ -20,7 +20,10 @@
 ## URI/URL detection and external program launching utilities.
 ## Used by the `gf` command to open URIs under the cursor.
 
-import std/[options, osproc, sequtils, strutils, unicode, uri]
+import std/[options, sequtils, strutils, unicode, uri]
+
+when not defined(moe.embedded):
```
fox0430: `git_diff.nim:25` and `types/git_diff_types.nim:26` (the latter while calling itself the lightweight types module that keeps callers from transitively pulling in the pipeline implementation) import it unconditionally, and a build confirms the `osproc` objects and `startProcessAuxSpawn__pureZosproc` survive into the linked binary.

This PR made the import conditional as well, so changing only the guard inside the proc leaves `startProcess` undeclared. With both changed, `gf` works in a Linux embedded build, as verified.

```nim
when not (defined(moe.embedded) and defined(windows)):
  import std/osproc
```

```nim
  # inside the proc
  when defined(moe.embedded) and defined(windows):
```

fox0430 (changes requested):

src/moepkg/uri_utils.nim
```diff
@@ -182,15 +187,19 @@ proc fileUriToPath*(uri: string): string =
...
-  let cmd = when defined(macosx): "open" else: "xdg-open"
-
-  try:
-    let p = startProcess(cmd, args = @[uri], options = {poUsePath})
-    p.close()
-    return true
-  except OSError:
-    return false
+  when defined(moe.embedded) and defined(windows):
+    discard uri
```
fox0430: `discard uri` is unnecessary code

fox0430 (changes requested):

src/moepkg/clipboard_embedded.nim
```diff
@@ -0,0 +1,123 @@
...
+    write
+
+  ClipboardError* = object of CatchableError
+
+const
+  WriteTimeoutMs = 10_000
+  EmbeddedClipboardError =
+    "System clipboard access is owned by the embedding frontend (-d:moe.embedded)"
+
+proc getClipboardCommand*(
```
fox0430: These 3 command tables are a verbatim copy of clipboard.nim:765-813, 853-864, but every sync entry point here returns EmbeddedClipboardError unconditionally, so nothing in an embedded build ever calls them - and test_clipboard.nim imports clipboard directly, so they aren't covered either. Suggest deleting them.

fox0430 (changes requested):

fox0430 (approved):
Thanks!

elcritch:
Thank you!

PR #3251, opened 2026-09-04
feat(highlight): add optional Matter syntax backend

elcritch:
## Summary

Add an optional Matter TextMate syntax backend while keeping builtin highlighting as the default. Enable it with `nimble --parser:declarative --features:matter build -d:release` (Nimble 0.24.1 or newer). The feature requires Matter >= 0.4.2; default dependency installation excludes Matter and its dedicated dependencies.

Grammars are supplied explicitly through `[Highlight] matterGrammarFiles` inside Moe's configuration directory, or as grammar text through the editor/buffer API. Moe does not bundle or download grammars. Runtime `backend = "matter"` selects the backend; languages without a valid supplied grammar fall back to builtin highlighting. Diff and Log retain their builtin lexers.

The backend integrates with progressive loading, budgeted incremental parsing, Unicode columns, per-line length caps, reserved words, Markdown block detection, and URI/LSP/diagnostic overlays. Review fixes preserve progressive caches for empty grammar configurations, reject cached tokenizer states from a different backend, avoid comparing grammar contents for identical references, and retry a sticky Matter failure once when a later edit starts a fresh reparse flight.

Includes backend and regression tests, a Matter-enabled CI job (including config-loader coverage), example configuration, and build documentation.

## Validation

- Latest dependency change: default dependency installation and build passed in an isolated checkout without workspace vendored dependencies.
- Matter feature installation/setup selected Matter 0.4.2, and the feature-enabled release build passed.
- Earlier review fixes: Matter adapter, incremental highlighting, edge-case, public API, builtin highlighting, and release build checks passed locally, including regressions for progressive caches, mismatched cached backends, and failure recovery.
- Whitespace checks passed. The config-loader suite has a previously observed unrelated macOS clipboard-tool expectation failure; Matter CI runs on Linux.

Tokenization uses a configurable soft 20ms per-line deadline. Timeout or grammar errors leave the affected line and subsequent lines plain for that parsing flight; a later edit can retry. Equivalence tests disable the wall-clock deadline for deterministic results.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -192,4 +192,5 @@ jobs:
         nim c -r -d:moe.matter -d:matterTimeLimitMs=0 -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight_matter.nim
         nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_matter_highlight_edges.nim
         nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight.nim
+        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlighter_api.nim
         nim c -d:release -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" src/moe.nim
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -159,3 +159,37 @@ jobs:
         # whole job a silent no-op. Fail loudly instead.
         command -v lasm
         xvfb-run env XDG_SESSION_TYPE=x11 nim c -r tests/test_lsp_e2e_lasm.nim
+
+  matter:
+    name: Matter syntax backend
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      TERM: xterm-256color
+    steps:
+    - uses: actions/checkout@v7
+    - uses: jiro4989/setup-nim-action@v2
+      with:
+        nim-version: 2.2.10
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Install dependencies
+      run: |
+        nimble install -d -y
+        nimble setup
+    # Matter 0.3.0's Nimble package omits its grammar archives. Use the exact
+    # locked source revision for build-time assets, never a moving branch.
+    - name: Fetch Matter grammar assets
+      uses: actions/checkout@v7
+      with:
+        repository: elcritch/matter
+        ref: 8db6f03d1028c42693c3a586ba3625e1b48e9589
+        path: deps/matter-assets
+        persist-credentials: false
+    - name: Build and test Matter backend
+      run: |
+        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_matter_backend.nim
+        # Disable wall-clock interruption for full-vs-incremental equivalence.
+        nim c -r -d:moe.matter -d:matterTimeLimitMs=0 -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight_matter.nim
+        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_matter_highlight_edges.nim
+        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight.nim
+        nim c -d:release -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" src/moe.nim
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -196,21 +196,12 @@ jobs:
       run: |
         nimble install -d -y
         nimble setup
-    # Matter 0.3.0's Nimble package omits its grammar archives. Use the exact
-    # locked source revision for build-time assets, never a moving branch.
-    - name: Fetch Matter grammar assets
-      uses: actions/checkout@v7
-      with:
-        repository: elcritch/matter
-        ref: 8db6f03d1028c42693c3a586ba3625e1b48e9589
-        path: deps/matter-assets
-        persist-credentials: false
     - name: Build and test Matter backend
       run: |
-        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_matter_backend.nim
+        nim c -r -d:moe.matter tests/test_matter_backend.nim
         # Disable wall-clock interruption for full-vs-incremental equivalence.
-        nim c -r -d:moe.matter -d:matterTimeLimitMs=0 -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight_matter.nim
-        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_matter_highlight_edges.nim
-        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlight.nim
-        nim c -r -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" tests/test_highlighter_api.nim
-        nim c -d:release -d:moe.matter -d:matterGrammarRoot="$PWD/deps/matter-assets" src/moe.nim
+        nim c -r -d:moe.matter -d:matterTimeLimitMs=0 tests/test_highlight_matter.nim
+        nim c -r -d:moe.matter tests/test_matter_highlight_edges.nim
+        nim c -r -d:moe.matter tests/test_highlight.nim
+        nim c -r -d:moe.matter tests/test_highlighter_api.nim
+        nim c -d:release -d:moe.matter src/moe.nim
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -192,6 +192,8 @@ jobs:
       with:
         nim-version: 2.2.10
         repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Update Nimble for revision-pinned dependencies
+      run: nimble install 'nimble@0.24.1' -y
     - name: Install dependencies
       run: |
         nimble install -d -y
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -196,8 +196,8 @@ jobs:
       run: nimble install 'nimble@0.24.1' -y
     - name: Install dependencies
       run: |
-        nimble install -d -y
-        nimble setup
+        nimble --parser:declarative --features:matter install -d -y
+        nimble --parser:declarative --features:matter setup
     - name: Build and test Matter backend
       run: |
         nim c -r -d:moe.matter tests/test_matter_backend.nim
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -206,4 +206,5 @@ jobs:
         nim c -r -d:features.moe.matter tests/test_matter_highlight_edges.nim
         nim c -r -d:features.moe.matter tests/test_highlight.nim
         nim c -r -d:features.moe.matter tests/test_highlighter_api.nim
+        nim c -r -d:features.moe.matter tests/test_config_loader.nim
         nim c -d:release -d:features.moe.matter src/moe.nim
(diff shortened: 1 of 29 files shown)


fox0430:
@elcritch 

Would it be difficult to load arbitrary grammar files at startup, like a config file, and use them for highlighting?

elcritch:
> @elcritch
> 
> Would it be difficult to load arbitrary grammar files at startup, like a config file, and use them for highlighting?

It should be possible. The trickiest part is making sure the grammars are available. I've added a common selection in [matter releases](https://github.com/elcritch/matter/releases/tag/v0.3.0).

I'm not sure of the state of this PR since I haven't reviewed it yet. Matter supports loading grammars, but there's also some static options I added earlier.

elcritch:
What would be your thoughts on the Moe's current syntax highlighters? Keep them as defaults / fallbacks and then make TM grammars opt in with the config files?

fox0430:
@elcritch 

> What would be your thoughts on the Moe's current syntax highlighters? Keep them as defaults / fallbacks and then make TM grammars opt in with the config files?

Okay. Let's proceed with the current approach for now.

elcritch:
> > What would be your thoughts on the Moe's current syntax highlighters? Keep them as defaults / fallbacks and then make TM grammars opt in with the config files?
> 
> Okay. Let's proceed with the current approach for now.

Well I'm not quite sure what the current approach is myself. 😆 

It's an area that I need to understand better. I think having the TM grammars be downloaded or possibly included with Moe installations.

However it looks like Moe's syntax layer is already flexible so it should be possible to configure it at runtime with whatever policies makes sense.

elcritch:
> @elcritch
> 
> Would it be difficult to load arbitrary grammar files at startup, like a config file, and use them for highlighting?

I've updated the PR to do this. It also allows embeders to pass the grammar files when creating a Moe buffer.

I've now wired it into Kosmo to start using it!

elcritch:
BTW, I updated Matter's language collection to increase the zip compression level and it went from ~3MB to 0.85MB! 

That's with I think all the languages that Moe uses. Kosmo bundles them directly in the binary and uses zippy to decompress at runtime. Something similar might be useful for Moe, but that's a separate PR.

Another avenue I'd like to add would be support for automatically downloading .vsx extensions from the open source vs code repo (not MS'es). Matter already handles parsing and extracting the TM grammars from such files! 

The only remaining part is adding searching of the open source vs code repository. I'm unsure of the networking aspect, but Kosmo uses Chronos for downloading assets, so an optional Chronos based downloader might be useful. Moe supports Chronos for async right? 🤔

fox0430:
@elcritch 

> BTW, I updated Matter's language collection to increase the zip compression level and it went from ~3MB to 0.85MB!
> 
> That's with I think all the languages that Moe uses. Kosmo bundles them directly in the binary and uses zippy to decompress at runtime. Something similar might be useful for Moe, but that's a separate PR.
> 
> Another avenue I'd like to add would be support for automatically downloading .vsx extensions from the open source vs code repo (not MS'es). Matter already handles parsing and extracting the TM grammars from such files!
> 
> The only remaining part is adding searching of the open source vs code repository. I'm unsure of the networking aspect, but Kosmo uses Chronos for downloading assets, so an optional Chronos based downloader might be useful. Moe supports Chronos for async right? 🤔

That's a great idea. Yes. Since moe uses chronos, it should be possible to perform network programming asynchronously as well.

fox0430:
@elcritch 

> BTW, I updated Matter's language collection to increase the zip compression level and it went from ~3MB to 0.85MB!

Couldn't we further reduce the size by using zstd instead of zip?

elcritch:
Finally got Terraform `.hcl` files to get syntax highlighting!

Note that some files can take a few hundred milliseconds to parse and highlight. I've moved Kosmo's syntax parsing off the main UI thread for markdown files. I'm not actually sure what it's doing now for Moe, but it feels instant so far.

Screenshot 2026-09-08 at 10 59 04 AM

src/moepkg/highlight_config.nim
```diff
@@ -26,14 +26,17 @@ import config
 import buffer/[core, highlight]
 
 proc applyHighlightCap*(buffer: TextBuffer, config: EditorConfig) =
-  ## Seed the per-line tokenization cap from config. Must run BEFORE `loadFile`,
+  ## Seed the backend and per-line cap from config. Must run BEFORE `loadFile`,
   ## which reads `maxHighlightLineLength` to cap the first chunk. Applying it
   ## afterwards would nil the freshly-seeded progressive-load cache and force a
   ## full synchronous reparse on open — the stall the cap exists to prevent.
+  when defined(moe.matter) or defined(features.moe.matter):
+    buffer.setMatterGrammarSet(config.highlight.matterGrammarSet)
```
fox0430, line 34: This nils the progressive-load cache on the split path - the exact reparse the docstring and `window_manager.nim` are trying to prevent.

`newTextBuffer` leaves `matterGrammarSet` nil (`buffer/core.nim:413` declares it with no initializer) while `newEditorConfig()` stores a non-nil empty set (`config.nim:304`), and `==` (`matter_backend.nim:73-74`) treats nil and non-nil as unequal. So `setMatterGrammarSet` fires on the first call for every buffer, clearing `incrementalHighlight` and resetting `uriScanParsedUpTo`. The backend assignment right below is a no-op here (buffer and config both default to `hbBuiltin`), so this fires even with `backend = "builtin"`.

`vsplit`/`hsplit` copy only `maxHighlightLineLength` before `loadFile`, then `registerSplitBuffer` (`editor_window.nim:157`) calls `applyHighlightConfig` - so `:vs big.nim` / `:sp big.nim` and the multi-file auto-split startup throw away the 1000-line chunk `loadFile` just built.

Either initialize the field in `newTextBuffer`, or make `==` treat nil as an empty set.

elcritch:
> @elcritch
> 
> > BTW, I updated Matter's language collection to increase the zip compression level and it went from ~3MB to 0.85MB!
> 
> Couldn't we further reduce the size by using zstd instead of zip?

Probably, but there's some diminishing returns. I tried with xz at its highest level and it was something like 10-15% smaller.

I stuck with zip since zippy let's Kosmo embed the decoder too. Especially for windows it's helpful. I'm not aware of zstd Nim implementation or good static wrapper, but I'd be willing to add whatever to Matter.

src/moepkg/highlight.nim
```diff
@@ -1531,7 +1713,7 @@ proc updateHighlightIncremental*(
     if reparseStart > 0 and reparseStart - 1 < incrHighlight.lineStates.states.len:
```
fox0430: `refreshStateCache` (`:1840-1844`) grows `states` with `setLen(min(lineCount, handoff))`, filling the band below with `TokenizerState()` when the flight starts below the load frontier. Its own comment says consumers in that band must gate on `parsedUpTo`, but this anchor selection only bounds-checks `reparseStart - 1 < states.len`. The default `backend` is `hbBuiltin`, and `initHighlightIncrementalFromStr` (`:1267`) dispatches on `initialState.backend` alone, so a Matter buffer falls into the builtin path. `isCodeBlockLine` reads the same band per line and takes the builtin arm there too.

Narrow to hit: it needs a Matter buffer where an edit well below `parsedUpTo` starts a long flight, plus a second edit landing in the hole while that flight is still in flight (the load tick that would truncate the hole is skipped meanwhile). It is also self-healing - the next load tick truncates and reparses - but until then colors come out builtin and the convergence check compares states across backends.

The hole fill itself predates this PR, but it was harmless while `TokenizerState` had no `backend`. This PR already replaces the four `TokenizerState()` fallbacks in this proc with `freshTokenizerState` and gates the loops at `:1727` and `:1810` on `hbBuiltin` - the cached read is the one seed source left out.

Gating on `initialState.backend != incrHighlight.backend` and falling back to `freshTokenizerState`, or carrying the backend into the hole fill, would close it.

.github/workflows/test.yaml
```diff
@@ -179,3 +179,31 @@ jobs:
...
+        nimble setup
+    - name: Build and test Matter backend
+      run: |
+        nim c -r -d:moe.matter tests/test_matter_backend.nim
+        # Disable wall-clock interruption for full-vs-incremental equivalence.
+        nim c -r -d:features.moe.matter -d:matterTimeLimitMs=0 tests/test_highlight_matter.nim
+        nim c -r -d:features.moe.matter tests/test_matter_highlight_edges.nim
+        nim c -r -d:features.moe.matter tests/test_highlight.nim
+        nim c -r -d:features.moe.matter tests/test_highlighter_api.nim
+        nim c -d:release -d:features.moe.matter src/moe.nim
```
fox0430, line 210: Please add `nim c -r -d:features.moe.matter tests/test_config_loader.nim`

src/moepkg/syntax/matter_backend.nim
```diff
@@ -0,0 +1,293 @@
...
+    else:
+      previous.grammar
+  if grammar.isNil:
+    result.nextState.failed = true
+    return
+  result.nextState.grammar = grammar
+  try:
+    let parsed =
+      grammar.tokenizeLine(sanitizeInvalidUtf8(line), previous.stack, timeLimitMs)
+    if parsed.stoppedEarly:
```
fox0430, line 274: `stoppedEarly` sets `failed = true`, `failed` is sticky (`:255-257` returns immediately), and the failed state gets cached into `lineStates`. If line 10 of a 2000-line C++ file exceeds `MatterTimeLimitMs` (default 20) on a loaded machine, lines 10-1999 render plain, and every later edit below line 10 resumes from that cached failure via `reparseStart = changedStartLine - 2`. The rewind loop at `:1727` is gated on `initialState.backend == hbBuiltin`, so Matter never rewinds out of it - recovery needs an edit above line 10 or a buffer reopen. It also breaks full-vs-incremental equivalence, which is why that test needs `-d:matterTimeLimitMs=0`.

Clearing `failed` at the start of a reparse, or retrying once per flight, would recover.

src/moepkg/syntax/matter_backend.nim
```diff
@@ -0,0 +1,293 @@
...
+    ## Completed state entering the next line. A failure is sticky so a slow
+    ## or invalid grammar cannot be retried on every subsequent line/frame.
+    grammar: Grammar
+    stack*: StateStack
+    failed*: bool
+
+proc `==`*(a, b: MatterLineState): bool =
+  a.failed == b.failed and a.grammar == b.grammar and a.stack == b.stack
+
+proc `==`*(a, b: MatterGrammarSet): bool =
```
fox0430, line 71: Minor: this compares whole grammar sources on every buffer open. `applyHighlightCap` calls `setMatterGrammarSet` for every buffer and this `==` guards it; runtime caches are excluded, but `a.sources` compares each source's full `content` - megabytes for a real C++ grammar - and it does so even when `a` and `b` are the same ref.

No impact with the default (empty) config, so this is cheap waste rather than a bug. An identity early-out (`cast[pointer]` compare, or path/len first, since `==` recurses) removes it.

fox0430 (changes requested):

elcritch:
Addressed the latest review in 411585d7:

- preserve progressive-load caches by treating nil and empty Matter grammar sets as equal
- avoid deep grammar-source comparisons when both buffers share the same grammar-set reference
- reject cached tokenizer states whose backend does not match the incremental cache
- retry a sticky Matter failure once when a later edit starts a fresh reparse flight
- run `test_config_loader.nim` with the Matter feature in CI

Added deterministic regressions for the progressive builtin cache, Matter cache-gap state, and sticky-failure recovery. The Matter backend, incremental highlighting, edge-case, public API, builtin highlighting, and release build checks pass locally.

moe.nimble
```diff
@@ -18,6 +18,8 @@ requires "stew >= 0.2.0"
 requires "editorconfig >= 0.1.1"
 requires "regex >= 0.26.1"
 requires "jsony >= 1.1.6"
+# Matter 0.3.0 engine; enabled by its direct define or Nimble feature. No grammars bundled.
+requires "https://github.com/elcritch/matter#333b79c75fa883a3ebb5ddd7d4fdbd7a11819b94"
```
fox0430: Please use `feature`

```
feature "matter":
  requires "https://github.com/elcritch/matter#333b79c75fa883a3ebb5ddd7d4fdbd7a11819b94"
```
We won't be able to build with old nimble... but we can avoid downloading unnecessary packages.

fox0430 (changes requested):

moe.nimble
elcritch: Addressed in 29bceb64. Matter is now declared inside `feature "matter"`, preserving the remote update to `>= 0.4.2`. I also removed Matter, reni, and zippy from the default lockfile so ordinary installs do not pull them in, and updated the Matter CI job and build instructions to enable the feature explicitly with the declarative parser.

Validated a default dependency install/build in an isolated checkout, plus Matter feature installation/setup and a feature-enabled release build using Matter 0.4.2. The PR description now reflects the current dependency and caller-supplied grammar model.

fox0430 (approved):
Great! Thank you!

PR #3252, opened 2026-09-05
Add buffer-local named marks

elcritch:
Normal mode had unnamed bookmarks but no Vim-style named positions, so operators could not target a saved location. This adds buffer-local marks `a` through `z`: `ma` sets a mark, `'a` jumps linewise, and `` `a `` jumps to the exact position. Mark motions work with operators, including `d'a`, and operator repeats retain the named-mark target.

Marks follow line and column edits and participate in undo, redo, transactions, and all three buffer backends. The existing bookmark bindings move from `mm`/`mn`/`mp`/`mc` to `gmm`/`gmn`/`gmp`/`gmc` so all lowercase mark names remain available.

Validation:

- `nimble build -y`
- `nim c -r tests/test_normal_handler.nim`
- `nim c -r tests/test_buffer_undo_redo.nim`
- `nim c -r tests/test_command_registry.nim`
- `nim c -r tests/test_key_bindings.nim`
- `nim c -r tests/test_howtouse_docs_sync.nim`

The full parallel runner could not execute locally because its GNU `timeout` dependency is unavailable on this macOS host; every attempted test exited 127 before compilation.


src/moepkg/key_bindings/normal_bindings.nim
```diff
@@ -53,10 +53,10 @@ const NormalBindings: seq[tuple[key, cmd: string]] = @[
...
   ("g ,", "changelist-next"),
   # Bookmarks
-  ("m m", "bookmark-toggle"),
-  ("m n", "bookmark-next"),
-  ("m p", "bookmark-prev"),
-  ("m c", "bookmark-clear"),
+  ("g m m", "bookmark-toggle"),
+  ("g m n", "bookmark-next"),
+  ("g m p", "bookmark-prev"),
+  ("g m c", "bookmark-clear"),
```
elcritch, line 59: This seems like a breaking change?

fox0430: Ok. No problem

src/moepkg/buffer/core.nim
```diff
@@ -897,6 +905,29 @@ proc bookmarkShiftCallback(b: TextBuffer, event: RowColRemapEvent) =
     event, b.adjustBookmarksForInsert, b.adjustBookmarksForDelete
   )
 
+proc namedMarkShiftCallback(b: TextBuffer, event: RowColRemapEvent) =
```
fox0430, line 935: `rrekMultiLine` only moves rows. `ckDeleteRange` has `preservesFirstRow: true`, so `row = firstAffectedRow + 1` and marks on the edited line are always excluded. The fixup at `core.nim:1342` only fires when `deleteEndPos.line > deleteStartPos.line`, and only for marks on `deleteEndPos.line`

Intra-line deletions need the same column shift as `rrekSingleLine`; multi-line deletions need the `core.nim:1342` fixup extended to the `startLine` side.

src/moepkg/buffer/undo.nim
```diff
@@ -111,6 +112,9 @@ proc undoChange(b: TextBuffer, change: BufferChange): Result[(), string] =
     # savedModifiedLines restore below would clobber their output.
     if change.kind notin {ckSnapshot, ckTransaction}:
       b.emitRowColRemapEvents(change, reverse = true, includeSideArrays = false)
+      b.namedMarks = default(typeof(b.namedMarks))
```
fox0430: `marksBefore` is restored wholesale, so marks on unrelated lines are rolled back too. Setting a mark pushes no undo entry, so "edit -> mark -> undo" loses the mark.

Vim keeps `z`. The restore should apply only to marks on the affected lines.

src/moepkg/buffer/core.nim
```diff
@@ -897,6 +905,29 @@ proc bookmarkShiftCallback(b: TextBuffer, event: RowColRemapEvent) =
...
+      of rrekSingleLine:
+        if pos.line == event.row and pos.column >= event.editCol:
+          pos.column = max(event.editCol, pos.column + event.colDelta)
+          pos.column = min(pos.column, max(0, event.lineCharLenAfter - 1))
+        mark = some(pos)
+      of rrekMultiLine:
+        let
+          delta = event.lastAffectedRowAfter - event.lastAffectedRowBefore
+          row = event.firstAffectedRow + ord(event.preservesFirstRow)
+        if delta < 0 and pos.line >= row and pos.line < row - delta:
```
fox0430, line 951: `joinLines` is `deleteLine` ×2 plus `insert`, so both `ckDeleteLine` entries fall in the drop range and are never remapped.

Vim keeps both on the joined line.

src/moepkg/buffer/undo.nim
```diff
@@ -538,6 +546,9 @@ proc redoChange(b: TextBuffer, change: BufferChange): Result[(), string] =
     # below overwrite the two per-line arrays so `includeSideArrays=false`.
     if change.kind notin {ckSnapshot, ckTransaction}:
       b.emitRowColRemapEvents(change, includeSideArrays = false)
+      b.namedMarks = default(typeof(b.namedMarks))
```
fox0430: The mirror image of the above, via the wholesale `marksAfter` restore.

```
ma on line 2, x on line 0, u, mb on line 1, redo -> a survives, b becomes none
```

src/moepkg/command_registry.nim
```diff
@@ -461,6 +461,42 @@ proc executeCommand*(
     # the command reached us — no prefix bookkeeping is needed here.
     let count = cmd.count
     case cmd.operatorType
+    of "mark-set", "mark-line", "mark-exact":
```
fox0430, line 464: The error path already clears it. The jump side (`'a` / `` `a ``) leaks on success and on `dest.isErr` alike.

```
"a then mb -> some('a') is still there
"c then 'b -> some('c') is still there
```

src/moepkg/buffer/core.nim
```diff
@@ -160,6 +162,7 @@ type
       ## collisions (undo -> different edit re-hitting savedSeq). Inner
       ## transaction changes carry 0; only the wrapper that lands on
       ## undoStack gets an id. 0 = initial state.
+    marksBefore*, marksAfter*: seq[tuple[name: char, pos: BufferPosition]]
```
fox0430: With even a single mark set, every edit allocates two `seq`s and holds them for the lifetime of an unbounded undo stack (~1.2 KB per entry with 26 marks). That runs against the direction of shrinking the side arrays with `CowSeq` / `SeqDelta`, so a delta representation seems worth considering.

fox0430 (changes requested):
More tests

- **Column tracking** - extend `followsEdits` with column cases: an intra-line delete (`ma` at column 11 of `"alpha beta gamma"` -> `dw` at column 0 → column 5) and a multi-line `deleteRange` with a mark left on the startLine side. `exactDelete` already asserts `` d`a `` at column precision, so the edit-tracking side deserves the same granularity.

- **Undo/redo preserve marks** - the same three-backend loop as `undoMarkDeletion`,
  for "edit -> `mz` on an unrelated line -> `u`" and "`ma` -> edit- > `u` -> `mb` -> redo". Each backend splits into ckSnapshot and non-ckSnapshot paths, so the loop earns its keep.

- **`J` preserves marks** - put a mark on each line before joining. It can be called directly at the `buffer` layer, so `tests/test_buffer*.nim` is fine too.

- **`pendingRegister` consumption** - `"a` then `mb` / `'b`, asserting `state.pendingInput.pendingRegister.isNone`. `block unsetAndInvalid` already covers the error path, so this sits right next to it.

- **Doc sync** - `test_help_keybinding_sync` guards only howtouse.md, which is exactly how features.md drifted here. Either bring the features.md table under AUTO-GEN, or at least assert that the bookmark keys are spelled `g m ...`.

src/moepkg/buffer/undo.nim
elcritch: Handled in 8c93a059. Undo entries now carry sparse before/after deltas only for mark names affected by the edit, so marks set after an edit on unrelated lines survive undo. Covered across GapBuffer, SqrtDecomp, and PieceTable.

src/moepkg/buffer/core.nim
elcritch: Handled in 8c93a059. joinLines remaps marks from both source lines onto the joined line, including whitespace trimming, and transaction deltas retain marks touched by intermediate operations. The regression also checks undo and redo.

src/moepkg/buffer/core.nim
elcritch: Handled in 8c93a059. Intra-line deleteRange now emits a column remap using the measured column delta, and multi-line ranges remap marks on the surviving start-line side. Added column-precision regressions for dw and multi-line deleteRange.

src/moepkg/buffer/core.nim
elcritch: Handled in 8c93a059. The two per-entry mark snapshots were replaced with one sparse delta containing only touched names. Edits that do not move a mark leave the delta empty and allocate no mark sequence.

src/moepkg/command_registry.nim
elcritch: Handled in 8c93a059. Every named-mark command now consumes pendingRegister before validation or destination lookup, covering successful mark set/jump and failed lookup paths.

src/moepkg/buffer/undo.nim
elcritch: Handled in 8c93a059. Redo applies only the sparse mark delta instead of restoring all marks wholesale, so marks set after undo survive. Covered across the same three backends.

src/moepkg/command_registry.nim
```diff
@@ -461,6 +461,41 @@ proc executeCommand*(
...
+      let op = ctx.state.pendingInput.pendingOperator
+      ctx.state.pendingInput.pendingOperator = none(PendingOperator)
+      ctx.state.pendingInput.pendingRegister = none(char)
```
fox0430: Moving the pendingRegister reset to the top of the branch also clears it for the operator path, so "qd'a / "qy'a no longer reach the named register. executeOperatorOnRange reads pendingInput.pendingRegister (operator_engine.nim:79-107) to route the text, and it already clears it afterwards (operator_engine.nim:148, 253, 331). Clearing it before applyOperatorToMark runs leaves only the unnamed register.

"qd'a  ->  before: register q = "one"   after: register q = empty
 
The leak I reported was on the jump side ('a / `a). Clearing it in the mark-set branch and just before recordJump covers that without  touching the operator path.

markCommandsConsumePendingRegister does not exercise an operator with a register prefix, so this slips through.

fox0430 (changes requested):

fox0430 (approved):
Thanks!

PR #3255, opened 2026-09-07
feat: support event-driven Git refresh for embedded frontends

elcritch:
This PR cleans up current git process handling as well as adding an embedding API to allow Kosmo to control the git process changes.

The git process handling issue is important since it appears to be a core cause of OS resource exhaustion in Kosmo. It's possible this might occur with Moe, but Kosmo also ran it's own git processes as well.

## Summary

- add an event-driven Git refresh mode and repository-change notifications for embedded hosts
- move branch lookups to shared asynchronous per-worktree cache entries and discard superseded results
- ensure cancelled and timed-out Git children are reaped and their resources are closed
- document embedding Git ownership and add regression coverage for events, worktrees, stale results, and cleanup

## Tests

- `nim c -r tests/test_git_refresh_events.nim`
- `nim c -r tests/test_editor_display.nim`


src/moepkg/git_cache.nim
```diff
@@ -135,26 +235,52 @@ proc refreshGitBranch*(gc: var GitCacheState, b: TextBuffer) =
...
     return
-
-  let branchResult = getGitBranch(filePath)
-  entry.path = filePath
-  entry.lastRefresh = now
-  entry.populated = true
-  entry.name = if branchResult.isErr: "" else: branchResult.get
-  gc.branchEntries[key] = entry
+  var repository = gc.repositories.getOrDefault(entry.repositoryPath)
+  let expired =
```
fox0430, line 288: `expired` reduces to `forced or not populated` in event-driven mode, and the four failure paths below all set `populated = true` with `forced` false. The only way back is a host `notifyGitRepositoryChanged` for that worktree - the sole writer of `repository.forced`.

Setting `forced = true` on those paths is not the fix: `expired` then stays true, so every tick starts another `rev-parse` - a retry loop under exactly the fd/process pressure this PR relieves. What's missing is a mode-independent failure backoff: a retry deadline on the entry, with the TTL applying to failures in both modes.

src/moepkg/git_cache.nim
```diff
@@ -57,30 +76,73 @@ proc setGitDiffRefreshInterval*(gc: var GitCacheState, ms: int64) =
...
   entry.populated = true
+  result = true
+
+proc reapBranch(entry: var GitRepositoryCacheEntry): bool =
+  if entry.pending.isNil:
+    return
+  try:
+    let exitCode = entry.pending.peekExitCode()
+    if exitCode == -1:
+      if (getMonoTime() - entry.started).inSeconds < 5:
```
fox0430, line 152: The 5s timeout. A slow first `rev-parse` (cold or large repo) leaves the branch blank until the host happens to send a repository notification.

src/moepkg/git_cache.nim
```diff
@@ -57,30 +76,73 @@ proc setGitDiffRefreshInterval*(gc: var GitCacheState, ms: int64) =
...
+      return
+    entry.name =
+      if exitCode == 0:
+        output.strip()
+      else:
+        ""
+    entry.lastRefresh = getMonoTime()
+    entry.populated = true
+    result = true
+  except CatchableError:
```
fox0430, line 169: `peekExitCode`/`readAll` raised. Same blank branch, same dead end.

src/moepkg/git_cache.nim
```diff
@@ -135,26 +235,52 @@ proc refreshGitBranch*(gc: var GitCacheState, b: TextBuffer) =
...
+    try:
+      repository.pending = startProcess(
+        "git",
+        args = [
+          "-C", entry.repositoryPath, "--no-optional-locks", "rev-parse",
+          "--abbrev-ref", "HEAD",
+        ],
+        options = {poUsePath, poStdErrToStdOut},
+      )
+    except CatchableError:
```
fox0430, line 307: `startProcess` raised. A transient `EAGAIN` blanks the branch permanently.

src/moepkg/types/git_cache_types.nim
```diff
@@ -44,13 +51,27 @@ type
 
   GitBranchCacheEntry* = object
     path*: string
+    repositoryPath*: string
+    name*: string
```
fox0430: This PR removed the only assignment to `GitBranchCacheEntry.name`. The remaining one (`git_cache.nim:124`) is on `GitRepositoryCacheEntry`, and the sole reader is the `else` fallback in `gitBranchName`, reachable only from tests that build the entry with an empty `repositoryPath` (`test_editor_display.nim:85`, `:98`, `:100`, `:114`; `test_statusline.nim:1497`, `:1512`, `:1523`). Those tests now assert on a path production cannot take - worth dropping the field and pointing them at `repositories`.

src/moepkg/git_cache.nim
```diff
@@ -180,6 +308,17 @@ proc evictGitCacheForBuffer*(gc: var GitCacheState, b: TextBuffer) =
       entry[].pending = none(GitDiffProcess)
   gc.diffEntries.del(key)
   gc.branchEntries.del(key)
+  var unused: seq[string]
```
fox0430, line 349: `used` is computed from `branchEntries` only, ignoring `GitDiffCacheEntry.repositoryPath`, while `documents/embedding-git.md:59` says the branch query is released when no other buffer *uses that worktree*. Costs one extra `rev-parse`; doc and code disagree.

src/moepkg/git_diff.nim
```diff
@@ -83,6 +83,27 @@ proc closeSafely(p: Process) =
...
+  if process.isNil:
+    return
+  let child = process
+  process = nil
+  try:
+    if child.peekExitCode() == PROCESS_RUNNING:
+      terminateProcess(child)
+      let deadline = getMonoTime() + initDuration(milliseconds = 100)
+      while child.peekExitCode() == PROCESS_RUNNING and getMonoTime() < deadline:
+        sleep(1)
```
fox0430, line 98: `releaseGitProcess` can spend ~100ms in `sleep(1)` on the editor thread, and it is reachable from buffer close, not just background ticks. Only bites when the child ignores SIGTERM, so probably fine as is.

src/moepkg/git_cache.nim
```diff
@@ -95,12 +157,19 @@ proc scheduleGitRefresh*(gc: var GitCacheState, b: TextBuffer) =
   let now = getMonoTime()
   let needsRefresh =
```
fox0430, line 195: Same gate on the diff side, with a milder outcome. When `startGitDiffFromBufferAsync` fails, `:178-181` sets `gitTracked = false` and `populated = true`; `needsRefresh` still covers `changeSeqAtRefresh` and `pathAtRefresh`, so an edit, save, reload or gutter toggle retries - but a buffer the user only reads keeps `gitTracked = false` indefinitely, and `editor_render_window.nim:1150` drops its gutter on that flag. The comment at `:179` ("count the failed start as an attempt so we don't retry every tick") was written when the TTL was unconditional and did the retrying.

fox0430 (changes requested):
`sourceBuffer` is now a strong ref to a `ref object`, and `closeRenameTargets` (`editor_lsp_rename.nim:48`) calls `deleteBufferAt` without evicting first — unlike the three paths that do (`editor_buffers.nim:260`, `:489`, `viewer_mode.nim:49`), two of which say so in a comment. Not reachable today (rename targets are never displayed, so they never get a diff entry), but the invariant has a hole.

src/moepkg/git_cache.nim
elcritch: Added a mode-independent `retryAfter` deadline in fdcf6c9c. Failed branch queries now back off for five seconds in both refresh modes, while successful queries still follow the selected refresh mode. Regression coverage exercises both periodic and event-driven modes.

src/moepkg/git_cache.nim
elcritch: Fixed in fdcf6c9c. A timed-out `rev-parse` now preserves the last branch value and schedules a retry after the five-second backoff, including in event-driven mode. The regression test covers the timeout path.

src/moepkg/git_cache.nim
elcritch: Fixed in fdcf6c9c. Exceptions from polling or reading now go through the same delayed-retry path and preserve the last published branch value.

src/moepkg/git_cache.nim
elcritch: Fixed in fdcf6c9c. A failed `startProcess` records a retry deadline instead of becoming permanently populated. The retry test injects a spawn failure and verifies that it neither loops every tick nor gets stuck in either refresh mode.

src/moepkg/types/git_cache_types.nim
elcritch: Fixed in fdcf6c9c. Removed `GitBranchCacheEntry.name`, made `gitBranchName` read only from repository entries, and updated the display and status-line tests to model production repository-backed state.

src/moepkg/git_cache.nim
elcritch: Fixed in fdcf6c9c. Repository eviction now checks both branch entries and diff entries, and the new regression keeps a shared branch process alive while a diff-only buffer still uses the worktree.

src/moepkg/git_diff.nim
elcritch: Agreed. I left the bounded 100 ms termination grace unchanged. The existing cleanup regressions cover both ordinary termination and a child that ignores SIGTERM, and they pass.

src/moepkg/git_cache.nim
elcritch: Fixed in fdcf6c9c. Failed diff starts now set `retryAfter` using the configured Git interval, which applies in both refresh modes. The regression verifies the backoff prevents per-tick retries and that the diff recovers after the deadline.

documents/embedding-git.md
```diff
@@ -0,0 +1,65 @@
...
+Worktree discovery walks ancestor directories for `.git` files or directories,
+resolving directory symlinks. Normal repositories, linked worktrees and submodules
+are supported; custom repositories selected only through `GIT_DIR`/`GIT_WORK_TREE`
+environment overrides are not part of this discovery contract.
+
+Closing a buffer releases its pending diff and its branch query if no other
+buffer uses that worktree. Editor shutdown releases all outstanding Git work.
+Cancellation and timeout allow a short termination grace period, then kill and
+reap the child before closing handles and removing scratch files. Call
+`releaseExternalResources()` when disposing of an embedded editor.
```
fox0430, line 65: Please add `handler.releaseExternalResources` to the export list (frontend.nim).

src/moepkg/git_cache.nim
```diff
@@ -122,11 +199,42 @@ proc requestGitRefresh*(gc: var GitCacheState, b: TextBuffer) =
...
   entry.forced = true
+  entry.pendingDiffInfo = none(GitDiffInfo)
   gc.diffEntries[key] = entry
 
+proc setGitRefreshMode*(gc: var GitCacheState, mode: GitRefreshMode) =
+  ## Select how elapsed time affects scheduling. Explicit requests and edits
+  ## continue to refresh in either mode; pending children are still reaped.
+  gc.refreshMode = mode
+
+proc notifyGitRepositoryChanged*(gc: var GitCacheState, rootPath: string) =
```
fox0430, line 240: A scoped notification only matches entries whose already-cached `repositoryPath` equals the notified root (or is empty), so it can never re-key an entry onto a newly created nested repository - which the doc explicitly promises ("The host must notify … repository creation/removal").

Repro: repo at `R` with branch `outerbranch`, buffer `R/sub/f.txt`; then `git init` + commit in `R/sub` and `notifyGitRepositoryChanged(R/sub)` -> `diffEntries[b].forced == false`, branch still `outerbranch`, `branchEntries[b].repositoryPath` still `R`. Under `grmEventDriven` there is no TTL to recover, so `git init sub/` or `git submodule add` leaves both the branch and the diff gutter wrong permanently unless the host falls back to `notifyGitRepositoryChanged("")`.

Suggest also invalidating entries whose `repositoryPath` is a proper ancestor of `root` (and the branch entries under it).

fox0430 (changes requested):

documents/embedding-git.md
elcritch: Fixed in 8d3fb44a: frontend.nim now exports handler.releaseExternalResources, and tools/embedded_smoke.nim calls it to cover the embedding facade export. All 14 PR CI checks pass.

src/moepkg/git_cache.nim
elcritch: Fixed in 8d3fb44a: scoped notifications now invalidate cached repository entries when the notified root is the same as or nested below the cached worktree, allowing branch and diff entries to rediscover and re-key. Regression coverage includes nested repository creation/removal and sibling isolation. All 14 PR CI checks pass.

elcritch:
CI is green on 8d3fb44a. The remaining review feedback is addressed: the embedding facade exports releaseExternalResources; scoped repository notifications re-key cached branch/diff entries for nested repository creation/removal; and the earlier LSP rename cleanup fix evicts Git cache entries before deleting buffers, covered by test_editor_lsp_rename.

fox0430 (approved):
Thanks!

PR #3302, opened 2026-09-22
feat: add host hooks for popups and help

elcritch:
## Summary
- Add host-controlled popup completion rendering for embedded integrations.
- Allow the host to take over `:help` via a consumable request flag.
- Route Config `:q` variants through Config cleanup so embedded editors are not closed.

These hooks support Kosmo’s native popup and help panels.


src/moepkg/command_handlers/result_processor.nim
```diff
@@ -663,11 +663,15 @@ proc executeCommandOverlay*(e: Editor, commandText: string): bool =
...
   # 2. dispatch
   let activeBuffer = e.activeBuffer()
+  let commandMode = e.currentMode
   let isShared = e.isBufferShared(activeBuffer)
   let otherModifiedCount = e.modifiedBufferCountExcept(activeBuffer)
-  let r = e.handlerManager.handleCommandMode(
+  var r = e.handlerManager.handleCommandMode(
     activeBuffer, commandText, isShared, e.activeWindow.cursor.line, otherModifiedCount
   )
+  if commandMode == EditorMode.Config and
```
fox0430, line 672: The guard matches raw text, but `parseAndExecute` already decided three lines above. `parser.nim:113` reads only `parts[0]`, so `:q foo` and `: q` are `claQuit` yet match no literal: they fall through to `executeQuit`'s `hrCloseWindow`, which leaves the scratch buffer in `e.buffers` and - with Config as the only window - quits the editor, the case this guard is for. It drifts the other way too: a user alias (`exit = { command = "quit" }`) is not intercepted, a disabled `q` still is.

Keying off the dispatch result covers all of these, drops the `strutils` import, and folds `:close` correctly as well. Both kinds are `hrgExitAndResync`, so the teardown below is unchanged:

```nim
if commandMode == EditorMode.Config and r.kind == hrCloseWindow:
  r = HandlerResult(kind: hrConfigQuit)
```

`test_viewer_enter_exit_roundtrip.nim:779` calls `processResult` directly and never reaches the guard, so a test driving `executeCommandOverlay(e, ":q")` and `":q foo"` would be worth adding.

fox0430 (changes requested):

